PR #14, opened 2016-01-18
[Issue 3] - Added support for https

poplav:
(c) Copyright IBM Corp. 2016

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,14 @@ help:
 build:
 	@docker build -t $(DASHBOARD_IMAGE_NAME) .
 
+gen-certs:
+	@mkdir -p certs ; \
+		cd certs ; \
+		openssl genrsa -des3 -out server.enc.key 1024 ; \
+		openssl req -new -key server.enc.key -out server.csr ; \
+		openssl rsa -in server.enc.key -out server.key ; \
+		openssl x509 -req -days 365 -in server.csr -signkey server.key -out server.crt
+
 run: CMD?=
 run: | build run-kernel-gateway
 	@docker run -it --rm \
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ KG_CONTAINER_NAME=kernel_gateway
 help:
 	@echo 'Make commands:'
 	@echo '             build - builds Docker image for dashboard proxy app'
+	@echo '         gen-certs - generate HTTPS key and certificate files'
 	@echo '               run - runs the dashboard proxy and kernel gateway containers'
 	@echo '         run-debug - like `run` but with node network logging enabled'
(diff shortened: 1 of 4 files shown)


bin/www
```diff
@@ -19,10 +22,27 @@ var port = normalizePort(process.env.PORT || '3000');
...
+    fs.existsSync(key_file_location) && fs.existsSync(cert_file_location)) {
+    var options = {
+       key: fs.readFileSync(key_file_location),
+       cert: fs.readFileSync(cert_file_location)
+    };
+    server = https.createServer(options, app);
+    debug('Using, HTTPS_KEY_FILE = ' + key_file_location);
+    debug('Using, HTTPS_CERT_FILE = ' + cert_file_location);
+    debug('Server set to HTTPS mode');
+} else {
```
parente: If the vars are  set, but the paths are invalid, I think an exception should be thrown. Otherwise, a user might *think* he/she has things setup to be secure, but wind up running an insecure HTTP server instead

Makefile
```diff
@@ -17,6 +17,14 @@ help:
 build:
 	@docker build -t $(DASHBOARD_IMAGE_NAME) .
 
+gen-certs:
+	@mkdir -p certs ; \
+		cd certs ; \
```
parente: Probably want to use && here so that if one step fails, the rest don't run.

Makefile
```diff
@@ -17,6 +17,14 @@ help:
 build:
 	@docker build -t $(DASHBOARD_IMAGE_NAME) .
 
+gen-certs:
```
jhpedemonte: Add this to the `help` target.

Makefile
```diff
@@ -17,6 +17,14 @@ help:
...
+	@mkdir -p certs ; \
+		cd certs ; \
+		openssl genrsa -des3 -out server.enc.key 1024 ; \
```
jhpedemonte: Add to readme that `openssl` is a prereq for the `gen-certs` target.

README.md
```diff
@@ -26,4 +26,8 @@ This project uses [Node.js](nodejs.org), [npm](npmjs.com) and [gulp](http://gulp
 
 ## Security
 
-The notebook code is never made available to the client -- it is only run in our proxy server. Execution request messages from the client which contain code are ignored.
+* The notebook code is never made available to the client -- it is only run in our proxy server. Execution request messages from the client which contain code are ignored.
+* To enable HTTPS:
+    1. Generate the key and certificate files by running `make gen-certs`
```
parente: The Makefile generates a self-signed certificate which is fine for dev, but not OK for production. You should point this out but also be sure to say that users can bring their own real cert/key.

PR #16, opened 2016-01-21
[Issue 3] - Redirect http server requests to https, when in https mode.

poplav:
(c) Copyright IBM Corp. 2016

### DIFF
--- a/Dockerfile.proxy
+++ b/Dockerfile.proxy
@@ -43,8 +43,8 @@ RUN npm run build
 # always expose server on all interfaces in a container
 ENV IP 0.0.0.0
 
-# expose the default express http/https port (3000) and the node debugger port (8080)
-EXPOSE 3000 8080
+# expose the default express http/https port (3000/3001) and the node debugger port (8080)
+EXPOSE 3000 3001 8080
 
 # default to running npm and the commands in package.json
 ENTRYPOINT ["npm", "run"]
--- a/Makefile
+++ b/Makefile
@@ -41,11 +41,16 @@ certs/server.pem:
 certs: certs/server.pem
 
 # Targets for running the nodejs app and kernel gateway in containers
+run: HTTP_PORT?=3000
+run: HTTPS_PORT?=3001
 run: | build run-kernel-gateway
 	@docker run -it --rm \
 		--name $(DASHBOARD_CONTAINER_NAME) \
-		-p 9700:3000 \
+		-p $(HTTP_PORT):$(HTTP_PORT) \
+		-p $(HTTPS_PORT):$(HTTPS_PORT) \
 		-p 9711:8080 \
+		-e PORT=$(HTTP_PORT) \
+		-e HTTPS_PORT=$(HTTPS_PORT) \
 		-e HTTPS_KEY_FILE=$(HTTPS_KEY_FILE) \
 		-e HTTPS_CERT_FILE=$(HTTPS_CERT_FILE) \
 		-e KERNEL_GATEWAY_URL=http://$(KG_CONTAINER_NAME):8888 \
--- a/bin/www
+++ b/bin/www
@@ -19,6 +19,7 @@ var config = require('../app/config');
  */
 
 var port = normalizePort(config.get('PORT') || 3000);
+var https_port = normalizePort(config.get('HTTPS_PORT') || 3001);
 var ip = config.get('IP') || '127.0.0.1';
 app.set('port', port);
 app.set('ip', ip);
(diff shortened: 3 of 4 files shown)


jhpedemonte:
This needs to be updated due to latest changes.

Makefile
```diff
@@ -33,6 +33,8 @@ run: | build run-kernel-gateway
 	@docker run -it --rm \
 		--name $(DASHBOARD_CONTAINER_NAME) \
 		-p 9700:3000 \
+		-p 9701:3001 \
+		-e HTTPS_HOST_PORT=9701 \
```
jhpedemonte: Is there a better way to do this? I think this is used to do the redirect, where it has to be the **external** port rather than the **internal** one.

So a user would specify one port in `config.json` and another port for Docker? What happens when they're not using Docker? The code is still looking for `HTTPS_HOST_PORT`.

bin/www
```diff
@@ -18,8 +18,8 @@ var nconf = require('nconf');
  * Get port from environment and store in Express.
  */
 
-var port = normalizePort(process.env.PORT || '3000');
-app.set('port', port);
+var http_port = normalizePort(process.env.PORT || '3000');
+var https_port = normalizePort(process.env.PORT || '3001');
```
jhpedemonte: We'll need a different env var for this. I suggest `HTTPS_PORT` for now.

bin/www
```diff
@@ -40,12 +40,19 @@ if (key_file_location !== undefined && cert_file_location !== undefined) {
        key: fs.readFileSync(key_file_location),
        cert: fs.readFileSync(cert_file_location)
     };
-    server = https.createServer(options, app);
+    server = https.createServer(options, app).listen(https_port);
+
+    http.createServer(function (req, res) {
+        var url_location = "https://" + req.headers['host'].split(":")[0] + ":" + nconf.get('HTTPS_HOST_PORT') + req.url;
```
jhpedemonte: Use `config.get('HTTPS_HOST_PORT')` instead (new way) and move this up with the other 2 port declarations.

Also, should this only be an env var, or should it also reside in config.json? If the latter, then we'd have 3 different ports specified in our config. But since this is only used for Docker purposes, maybe we can just use it as an env var for now.

Also, it should default to the value of HTTPS_PORT if not set.

bin/www
```diff
@@ -40,12 +40,19 @@ if (key_file_location !== undefined && cert_file_location !== undefined) {
...
+        res.writeHead(301, { "Location": url_location});
+        res.end();
```
jhpedemonte: I think it would be cleaner to rewrite above 2 lines as `res.redirect(301, url_location)`

Makefile
```diff
@@ -7,6 +7,8 @@ DASHBOARD_CONTAINER_NAME=dashboard-proxy
 DASHBOARD_IMAGE_NAME=jupyter-incubator/$(DASHBOARD_CONTAINER_NAME)
 KG_IMAGE_NAME=jupyter-incubator/kernel-gateway-extras
 KG_CONTAINER_NAME=kernel-gateway
+HTTP_PORT=3000
+HTTPS_PORT=3001
```
jhpedemonte: Should this be `?=`, so it can be overridden? I'm not strong in the shell-fu, but I think that's how it works.

parente: Yes, in make. Also, if they're only used in one target, put them on that target only. (I don't have the full source in front of me at the moment to check.)

bin/www
```diff
@@ -42,12 +43,19 @@ if (key_file_location || cert_file_location) {
        key: fs.readFileSync(key_file_location),
        cert: fs.readFileSync(cert_file_location)
     };
-    server = https.createServer(options, app);
+    server = https.createServer(options, app).listen(https_port, ip);
+
+    http.createServer(function (req, res) {
+        var url_location = "https://" + req.headers['host'].split(":")[0] + ":" + https_port + req.url;
+        res.writeHead(301, { "Location": url_location});
+        res.end();
```
jhpedemonte: Replace above 2 lines with `res.redirect(301, url_location)` instead. It better conveys what's going on in this code.

Dockerfile.proxy
```diff
@@ -43,8 +43,8 @@ RUN npm run build
 # always expose server on all interfaces in a container
 ENV IP 0.0.0.0
 
-# expose the default express http/https port (3000) and the node debugger port (8080)
-EXPOSE 3000 8080
+# expose the default express http/https port (3000/3001) and the node debugger port (8080)
+EXPOSE 3000 3001 8080
```
jhpedemonte: If ports are set to something other than 3000/3001, how will this work? Do we need to add a specific line to our makefile where we run this Docker container? Or does `-p` automatically do the `EXPOSE` as well?

parente: You don't have to name every port you want to EXPOSE. These are only named so that `-P` can work and automatically expose every declared port on a random external port.

I would leave these as is as the preferred defaults.

Makefile
```diff
@@ -41,10 +41,13 @@ certs/server.pem:
...
+run: HTTPS_PORT?=3001
 run: | build run-kernel-gateway
 	@docker run -it --rm \
 		--name $(DASHBOARD_CONTAINER_NAME) \
-		-p 9700:3000 \
+		-p $(HTTP_PORT):$(HTTP_PORT) \
+		-p $(HTTPS_PORT):$(HTTPS_PORT) \
 		-p 9711:8080 \
 		-e HTTPS_KEY_FILE=$(HTTPS_KEY_FILE) \
 		-e HTTPS_CERT_FILE=$(HTTPS_CERT_FILE) \
```
jhpedemonte: Need to define these ports within the Docker env:

```
		-e PORT=$(HTTP_PORT) \
		-e HTTPS_PORT=$(HTTPS_PORT) \
```

jhpedemonte:
Latest looks good. Tested with default and custom ports. Will merge when travis tests pass.

PR #17, opened 2016-01-22
Project automation and instruction updates

parente:
* README intro
* Simpler Dockerfiles (no more root npm!) that build faster (debug tools further up)
* More explicit version tagging


parente:
Subsumed by #19

Issue #18, opened 2016-01-22
Doc and automation


parente:
* [x] All params are nconf
* [x] Expose IP interface as setting (default to 127.0.0.1)
* [x] All params are documented in config.json and shown in make help
* [x] Complete README sections
* [x] make test within docker container
* [x] Make cert targets real targets in makefile

PR #19, opened 2016-01-22
Project automation and instruction updates

parente:
Fixes issue #18.


Dockerfile.kernel
```diff
@@ -1,25 +1,18 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-#FROM jupyter/minimal-kernel
 FROM jupyter/minimal-kernel:9cb6443840b2
```
jhpedemonte: Does this need to be updated yet?

Dockerfile.kernel
```diff
@@ -1,25 +1,18 @@
...
 
 USER root
 
 # install latest version of node & npm (this and git clone also requires ca-certs)
-RUN apt-get install -yq --no-install-recommends ca-certificates curl
+RUN apt-get update && \
+    apt-get install -yq --no-install-recommends ca-certificates curl
 RUN curl -sL https://deb.nodesource.com/setup_5.x | bash -
+RUN apt-get update && \
```
jhpedemonte: Why run the update twice?

Makefile
```diff
@@ -1,49 +1,62 @@
...
-		openssl rsa -in server.enc.key -out server.key && \
-		openssl x509 -req -days 365 -in server.csr -signkey server.key -out server.crt
+certs/server.pem:
+	@mkdir -p certs
+	@openssl req -new \
+		-newkey rsa:2048 \
+		-days 365 \
+		-nodes -x509 \
+		-subj '/C=XX/ST=XX/L=XX/O=generated/CN=generated' \
+    -keyout $@ \
```
jhpedemonte: Indentation issue

Makefile
```diff
@@ -75,10 +94,10 @@ dev-install: _dev-install-ipywidgets
 	npm install
 	npm run bower
 
-dev: KG_IP?=192.168.99.100
+dev: KG_IP?=$$(docker-machine ip $$(docker-machine active))
```
jhpedemonte: This makes it dependent on docker-machine. Do we want to do that?

Dockerfile.kernel
parente: It's always good to pin to a version regardless. docker-stacks change a lot and sometimes in backward incompatible ways.

Dockerfile.kernel
parente: Docker best practice: the prior line might be cached from weeks ago, but you changed the second line. Ubuntu / Debian like to rev packages with security fixes by taking them off the inet. When that happens, and you haven't run apt-get update again, you fail.

Makefile
parente: Will fix.

Makefile
parente: The  default becomes dependent on docker machine. But if user sets KG_IP, that code never runs.  (It's late evaled).

PR #22, opened 2016-01-22
[issue: 12] Enable travis

### LINKED ISSUE #12: Enable travis
See title.

parente:
Fixes #12.

Should be merged after #19 since it's based on it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 
 node_js:
-    - "4"
+    - "5"
 
 env:
     - CXX=g++-4.8
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+language: node_js
+
+node_js:
+    - "4.1"
+
+before_install:
+    - make _dev-install-ipywidgets
+
+before_script:
+    - npm install -g gulp
+
+notifications:
+  email:
+    on_success: change
+    on_failure: always
+  slack: cloudet:rMfRKhvsbZuCIRZpTct3kBI4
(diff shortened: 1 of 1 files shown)


jhpedemonte:
@parente said:
> whelp, travis has defeated me
> https://travis-ci.org/jupyter-incubator/dashboards_nodejs_app/builds/104186360
> everything installs properly, including debug, but it fails finding debug

First thing I noticed from that log is the **node** and **npm** versions. Starts out as node version **0.10.x**, but later is updated to node version **4.2.6** and npm version **2.4.12**. At one point, I believe we had a problem with node/npm which went away when we update both to the newer version (5.0.x and 3.3.x, respectively). Maybe update `node_js` to "5" in `.travis.yml`.

Were you able to check the travis "image" to see if `debug` was in `/node_modules`? Was there anything in `/ext/ipywidgets/ipywidgets/node_modules`?

Searching for the error, only mention of relevance I found was [this commit](https://github.com/yyx990803/pod/commit/80b533bbf3d12fdcafa43fe9d76ec1f5aab56dc4). We don't explicitly include `debug` as a dependency (must be 1 or more layers down). But maybe we need to?

parente:
Node 5 FTW! I had been trying 5 last Friday, but never could get Travis to install it properly. It would get part way through the nvm download, fail, and go back to 0.10. So I went to 4.x which seemed to always download and install properly. Apparently the 5 problem was just a blip (but shows how well this might work on travis in the future :/ ).

parente:
Tests ran fine twice. Merging.

PR #24, opened 2016-01-25
Add make targets for running tmpnb containers

jtyberg:
Paired with @nitind to create some make targets that will run/kill tmpnb containers (proxy and pool).  Currently, the targets can be used to launch tmpnb separately from kernel_gateway container.  

To run tmpnb:

`make run-tmpnb TOKEN=mytoken`

To kill:

`make kill-tmpnb`

Have not changed the `make run` target -- it still runs kernel_gateway container.  We can look at switching to tmpnb (instead of kernel_gateway) once the node app has more support for running against tmpnb.

One final note: had to switch the ENTRYPOINT and CMD args in Dockerfile.kernel due to tmpnb always running command using `/bin/sh -c` (see https://github.com/jupyter/tmpnb/issues/209).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,10 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-.PHONY: help build certs run run-debug run-logging run-kernel-gateway kill dev-install dev debug _dev-install-ipywidgets
+.PHONY: help build certs \
+	run run-debug run-logging run-kernel-gateway \
+	run-tmpnb tmpnb-proxy tmpnb-proxy\
+	kill dev-install dev debug _dev-install-ipywidgets
 
 DASHBOARD_CONTAINER_NAME=dashboard-proxy
 DASHBOARD_IMAGE_NAME=jupyter-incubator/$(DASHBOARD_CONTAINER_NAME)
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -70,7 +73,66 @@ run-kernel-gateway:
...
 
-kill:
+###### tmpnb
+
+MAX_LOG_SIZE:=50m
+MAX_LOG_ROLLOVER:=10
+
+token-check:
+	@test -n "$(TOKEN)" || \
+		(echo "ERROR: TOKEN not defined (make help)"; exit 1)
```
jhpedemonte: Is "TOKEN" name too generic for this usage? Maybe rename to something like `TMPNB_TOKEN`? Or is this token going to be used elsewhere?

jtyberg: Good point.  I'll change it to `TMPNB_PROXY_AUTH_TOKEN` because I think we'll end up with others, like `TMPNB_API_AUTH_TOKEN`.

parente:
Once it goes green, I'll try it locally since travis doesn't test any of this.

jtyberg:
I reverted the change that switched ENTRYPOINT and CMD in `Dockerfile.kernel` to match the latest what the latest minimal-kernel docker stack is doing.

Makefile
```diff
@@ -70,7 +73,66 @@ run-kernel-gateway:
...
+			--max_dock_workers=16 \
+			--mem_limit=$(MEMORY_LIMIT) \
+			--redirect_uri=/api \
+			--command='jupyter kernelgateway \
+				--KernelGatewayApp.log_level=DEBUG \
+				--KernelGatewayApp.ip=0.0.0.0 \
+				--KernelGatewayApp.port={port} \
+				--KernelGatewayApp.base_url={base_path}'
+
+run-tmpnb: tmpnb-proxy tmpnb-pool
```
parente: Doc in makefile help.

parente: And maybe README too?

parente:
```
~/p/j/dashboards_nodejs_app ❯❯❯ make run-tmpnb                                                          jtyberg-tmpnb
ERROR: TMPNB_PROXY_AUTH_TOKEN not defined (make help)
```

but there's nothing in help about it. Doc that too.

jtyberg: Yes, that was rude.

Added.

Issue #20, opened 2016-01-22
Allow dashboard server to send auth headers to kernel gateway


parente:
1. ~~tmpnb might be configured to require an auth token to request a KG server~~
2. KG server might be configured to require an auth token to request a kernel

Assume both of these keys can be set once at server start time via nconf, and just pass them along where appropriate. The keys should NOT be exposed to the frontend client.

* [x] Pass auth token to kernel gateway
* ~~Pass auth token to tmpnb~~
* [x] Tests?

parente:
Let's just worry about KG token to start. tmpnb is going to be harder work with than expected because of the route to instance mapping needed in #11. We can open another item later to address tmpnb.

jhpedemonte:
In that case, ready for review. @dalogsdon, can you take a look?

parente:
Fixed by PR #26

PR #26, opened 2016-01-25
Pass along auth token for kernel gateway

jhpedemonte:
Ref #20


jhpedemonte:
Only tested with kernel gateway. Will test with tmpnb.

routes/api.js
```diff
@@ -10,13 +10,14 @@ var config = require('../app/config');
 var Promise = require('es6-promise').Promise;
 var url = require('url');
 
-var KERNEL_GATEWAY_URL = config.get('KERNEL_GATEWAY_URL');
+var kgUrl = config.get('KERNEL_GATEWAY_URL');
+var kgAuthToken = config.get('KG_AUTH_TOKEN');
```
parente: Set a false-y default (blank string or null) in `config.json` which is serving as documentation of all of the defaults and optional settings when you run `make help`.

jhpedemonte: Done.

parente:
I'll have a look at this once it's rebased.

jhpedemonte:
OK, got things working again after the merge. Might be able to do further cleanup and remove duplication of code with a `define` or two. Suggestions welcome.

Makefile
```diff
@@ -167,21 +172,43 @@ run-tmpnb-pool: token-check
...
+integration-test: IP?=$$(docker-machine ip $$(docker-machine active))
+integration-test: KG_PORT?=8888
+integration-test: KG_AUTH_TOKEN?=1a2b3c4d5e6f
+integration-test: | kill build run-kernel-gateway
+	$(DOCKER_APP) -d \
+		-e KERNEL_GATEWAY_URL=http://$(KG_CONTAINER_NAME):8888 \
+		-e KG_AUTH_TOKEN=$(KG_AUTH_TOKEN) \
+		--link $(KG_CONTAINER_NAME):$(KG_CONTAINER_NAME) \
+		$(DASHBOARD_IMAGE_NAME)
+	@echo 'Waiting 30 seconds for server to start...'
```
parente: Can probably shorten this quite a bit because there's another docker build that happens right after it.

Issue #15, opened 2016-01-20
POST handler for adding a notebook to the running instance


parente:
It would be handy if the node app supported `POST /notebooks` to add / overwrite an existing notebook with the same with a new one for quick and dirty deployment.

https://github.com/jupyter-incubator/dashboards_nodejs_app/blob/master/routes/index.js

Let's not worry yet about persisting these or PUT vs POST or overwrite vs error. I want to get a simple demonstration of how a bundler can be written for Jupyter Notebook server to post a simple notebook to this API to get it to show up as a dashboard instantly.

dalogsdon:
Discussed with @jtyberg. We'll go with a `multipart/form-data` handler to upload a file. It will upload to the `NOTEBOOKS_DIR` directory and overwrite any existing file.

The signature will follow the `GET /notebooks/*` API, so `POST /notebooks/*` will upload a file with optional subdirectory. Example: `POST /notebooks` will upload to the root data dir while `POST /notebooks/foobar` will create (if necessary) and upload to a folder called "foobar" in the data dir.

I have the initial implementation working. I still need to clean up a bit and finish a bit of error handling.

dalogsdon:
The form data key to use for the file is "file".

Example of calling from python (based on [this example](http://docs.python-requests.org/en/latest/user/quickstart/#post-a-multipart-encoded-file)):

```python
import requests
url = 'http://192.168.99.100/notebooks/newdashboard'
file = { 'file': open('my_dashboard.ipynb', 'rb') }
requests.post(url, files=file)
```

The uploaded dashboard filename will be based on the URL, not on the original file name. This way the same POST URL can be used in the browser to view the dashboard. In the python example above, the newly uploaded notebook can be viewed at `http://192.168.99.100/notebooks/newdashboard`

parente:
PR uses multer, so for future reference when we don't want to just dump the nb to local disk: https://github.com/expressjs/multer/blob/master/StorageEngine.md

parente:
Closed by PR #43.

PR #27, opened 2016-01-26
[Issue 15] POST /notebooks/* API

dalogsdon:
Also includes error handling updates.

:chicken:

Refs #15

### DIFF
--- a/routes/index.js
+++ b/routes/index.js
@@ -4,20 +4,10 @@
  */
 var debug = require('debug')('dashboard-proxy:server');
 var express = require('express');
-var multer = require('multer');
 var nbstore = require('../app/notebook-store');
-var uploadNb = require('../app/upload-notebook');
+var upload = require('../app/upload-notebook');
 
 var router = express.Router();
-
-// multer setup
-var storage = multer.diskStorage({
-    destination: uploadNb.destination,
-    filename: uploadNb.filename
-});
-var upload = multer({
-    storage: storage
-});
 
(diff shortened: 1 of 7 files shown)


parente:
LGTM but get another opinion from @jhpedemonte who's been living here longer than me.

routes/index.js
```diff
@@ -2,12 +2,23 @@
...
 
+// multer setup
+var storage = multer.diskStorage({
+    destination: uploadNb.destination,
+    filename: uploadNb.filename
+});
+var upload = multer({
+    storage: storage
+});
+
```
jhpedemonte: I would prefer that this code chunk (and require of `multer`, etc) be moved to ` app/upload-notebook.js`. Then in this file, we just require that file and set the `upload` object, to be used below. Makes the code cleaner and more modular.

dalogsdon: Yeah this is a leftover from some refactor. I'll move it over as suggested.

jhpedemonte:
Also, this is a candidate for integration test:

1. Upload notebook with 'phrase A'.
2. Query URL for that notebook and check that response has 'phrase A'.
3. Update notebook, now with 'phrase B'.
3. Query URL for notebook and check that response now has 'phrase B'.

dalogsdon:
Updated as per @jhpedemonte's comments.

Moved `multer` usage into the `upload-notebooks` module and did some basic form limit and file extension validation.

PR #33, opened 2016-01-27
Don't run build twice on `integration-test` target

jhpedemonte:
Decrease sleep time.
Add logging for major targets.


jhpedemonte:
cc @parente

PR #35, opened 2016-01-28
[Issue 32] - List notebooks.

poplav:
(c) Copyright IBM Corp. 2016


poplav:
screen shot 2016-01-28 at 11 59 43 am

views/index.handlebars
```diff
@@ -2,4 +2,27 @@
...
+<link rel="stylesheet" href="/css/style.css">
+
+{{> nav}}
+
+<div class="container">
+  <div class="col-md-4 col-md-offset-4">
+    <table class="table table-hover">
+      <thead>
+        <tr>
+          <th class="text-center">Available Notebooks</th>
```
parente: Maybe "Available Dashboards" not notebooks. End user might not even be aware that these are from notebooks.

parente:
Possible to add a unit test for this? Or integration test? Whichever makes sense?

views/index.handlebars
jhpedemonte: How about just "Dashboards"? Drop the "available".

jhpedemonte:
I think an integration test would work well here:

1. GET list and expect nothing.
2. Upload notebook of a certain name.
3. GET list again and expect that notebook name to be listed.

routes/index.js
```diff
@@ -15,9 +15,26 @@ router.get('/', function(req, res) {
...
-    res.render('index', {
-        title: 'Notebooks'
-    });
+    nbstore.getNotebooks().then(
+        function success(notebooks) {
+            //if notebook index exists redirect to it immediately
+            var notebookIndex = "index.ipynb"
+            if(notebooks.indexOf(notebookIndex) > -1) {
+                res.redirect('/notebooks/' + notebookIndex);
+            }
```
jhpedemonte: This only looks for "index" name, but not "Index" or "INDEX". Is that fine, or should we make this case-insensitive? I lean towards the latter.

routes/index.js
```diff
@@ -15,9 +15,26 @@ router.get('/', function(req, res) {
...
+            else {
+                //render a list of all notebooks
+                res.render('index', {
+                    username: req.session.username,
+                    notebooks: notebooks
+                });
+            }
+        },
```
jhpedemonte: If there is an "index" notebook, then we don't show the list. I think it would be good to have a "fallback" which always exists, such as the route `/list`.  If no "index", then `/` shows the same as `/list`. If there is an "index", it's shown at `/`, but `/list` is still accessible.  @parente, what do you think?

routes/index.js
jhpedemonte: Also, a `redirect` will change the client's URL bar. Might be better to just serve the file at `/` instead. Is that easy to do here?

routes/index.js
parente: Yes. Let's have a /list too.

poplav: Spoke with @jhpedemonte and in api.js we have a [check that the notebook path header matches](https://github.com/jupyter-incubator/dashboards_nodejs_app/blob/master/routes/api.js#L137) ```var matches = notebookPathHeader.match(/^\/notebooks\/(.*)$/);```  resulting in an error when we simply render on the index ```/```. It looks to be also used for storing  ```sessions[sessionId] = matches[1]; // store notebook path for later use``` The plan is to go with redirecting for now and open a subsequent issue where some refactoring would have to be done. @parente

app/notebook-store.js
```diff
@@ -186,5 +200,6 @@ module.exports = {
      */
     get: get,
     remove: remove,
-    upload: upload
+    upload: upload,
+    getNotebooks: getNotebooks
```
dalogsdon: Having both `get` and `getNotebooks` is ambiguous. Keeping with the nomenclature, we could use `list` here.

routes/index.js
```diff
@@ -9,16 +9,56 @@ var config = require('../app/config');
...
+/* GET / - index notebook or list of notebooks */
 router.get('/', function(req, res) {
-    res.redirect('/notebooks');
+    nbstore.getNotebooks().then(
+        function success(notebooks) {
+            //if notebook index exists redirect to it immediately
+            var notebookIndex = "index.ipynb"
+            //ensure notebook index search is case insensitive
+            var indexFound = -1;
+            for(i in notebooks) {
```
dalogsdon: Isn't `notebooks` an array? Typically `for..in` is used to iterate keys in an object. I would prefer either a `forEach` for its expressiveness or a normal `for` loop here if it is an array.

Additionally, it looks like you are creating an global variable `i`. I don't see any `var` declaration.

poplav:
Rebased and moved all test notebooks to test/resources.

views/index.handlebars
```diff
@@ -2,4 +2,27 @@
...
-<div>List of notebooks.</div>
+{{! Required for index }}
+<link rel='stylesheet' href='/components/bootstrap/css/bootstrap.min.css'>
+<link rel="stylesheet" href="/css/style.css">
+
+{{> nav}}
+
+<div class="container">
+  <div class="col-md-4 col-md-offset-4">
+    <table class="table table-hover">
```
dalogsdon: Reason for using a single-column table?

Could be more compact/semantic by using a list:
```html
<h1>Dashboards</h1>
<ul>
    {{each}}
        <li><a href="...">{{this}}</a></li>
    {{/each}}
</ul>
```

views/index.handlebars
```diff
@@ -2,4 +2,27 @@
     Copyright (c) Jupyter Development Team.
     Distributed under the terms of the Modified BSD License.
 }}
-<div>List of notebooks.</div>
```
dalogsdon: Should this template file be called "list.handlebars" since it is a list?

PR #38, opened 2016-01-29
Fix browserify build error in Docker

jhpedemonte:
Not sure why browserify fails in Docker, but decided
to just replace it all with webpack.


jhpedemonte:
Browserify is failing in Docker, but not locally (at least for me). Replaced with webpack, but there is still a build failure, now with `r.js` for some reason.

jhpedemonte:
The error in question:

```shell
> gulp build

[15:57:01] Using gulpfile ~/app/gulpfile.js
[15:57:01] Starting 'less'...
[15:57:01] Finished 'less' after 10 ms
[15:57:01] Starting 'webpack:components'...
[15:57:01] Starting 'copy:components'...
[15:57:02] Version: webpack 1.12.12
                Asset     Size  Chunks             Chunk Names
jupyter-js-widgets.js  1.12 MB       0  [emitted]  main
[15:57:03] Version: webpack 1.12.12
                 Asset    Size  Chunks             Chunk Names
jupyter-js-services.js  136 kB       0  [emitted]  main

ERROR in ./~/requirejs/bin/r.js
Module parse failed: /home/node/app/node_modules/requirejs/bin/r.js Line 1: Unexpected token ILLEGAL
You may need an appropriate loader to handle this file type.
| #!/usr/bin/env node
| /**
|  * @license r.js 2.1.22 Copyright (c) 2010-2015, The Dojo Foundation All Rights Reserved.
 @ ./~/jupyter-js-utils/lib/index.js 5:16-36
[15:57:03] Version: webpack 1.12.12
                    Asset    Size  Chunks             Chunk Names
jupyter-js-output-area.js  760 kB       0  [emitted]  main
[15:57:03] Finished 'webpack:components' after 1.72 s
[15:57:03] Finished 'copy:components' after 1.61 s
[15:57:03] Starting 'components'...
[15:57:03] Finished 'components' after 15 μs
[15:57:03] Starting 'build'...
[15:57:03] Finished 'build' after 5.58 μs
```

jhpedemonte:
Heh, this issue was caused by https://github.com/jupyter/jupyter-js-utils/commit/c36d0076399f7109f91cec7ace046d6ef8c50e08 (and related commits), which all went down yesterday! So a dependency of a dependency caused our build to break!

NPM! *(said in the style of Kirk's "KAAAAAHN!")*

jhpedemonte:
OK, got past build error (yay!). Now when running, 2 websocket connections are opened instead of 1 (WTF?!).

For some reason, `jupyter-js-services` is loading **dashboard.js** a 2nd time, kicking things off again. No idea why.

jhpedemonte:
I think the issues I'm seeing stem from changes for [handling require.js in `jupyter-js-utils`](https://github.com/jupyter/jupyter-js-utils/compare/84d334b5f8b309af20b847b3ad5c029e86c43f9b...master). When I use *webpack* to build `jupyter-js-services` (which pulls in `*-utils`), it builds *require.js* into the final file, even though we include it ourselves.

So during load of the dashboard, we load our *require.js*, which loads *dashboard.js* using the `data-main` attr. Later, `jupyter-js-services` loads its own *require.js* impl and goes through the initialization process again, processing `data-main` a second time and loading *dashboard.js* again.

I'm guessing there's some bit of webpack configuration that will prevent all this, but haven't found it yet. @blink1073, do you have any suggestions?

blink1073:
Argh, we were under the impression that it would use an existing instance of RequireJS and not clobber it.

blink1073:
cc @jdfreder

blink1073:
The problem might be that you are actually loading the `r.js` version using browserify and the other using WebPack.

jhpedemonte:
Well, I got rid of all browserify trying to fix this, so the build process shouldn't be loading `r.js` at all (unless webpack does somehow, or another build script).

blink1073:
This is the line I'm referring to: https://github.com/jrburke/r.js/blob/master/require.js#L182

jhpedemonte:
Instead of using `resolve.alias` in my webpack config, can I make use of `externals`? To say that require.js will be handled outside of `utils` and `services`?

blink1073:
Worth a shot.

jhpedemonte:
OK, `externals` is what I needed to use rather than `resolve.alias`, since we load our own require.js.

blink1073:
Great, sorry for the disruption!

parente:
Overlap issue aside, :shipit: 

screen shot 2016-01-29 at 4 13 19 pm

blink1073:
Just to follow up, I verified that using `resolve.alias` does *not* blow away the config when bundling (also tried with the CDN version of requirejs in the script tag).

index.html:
```
    <script src="require.js"></script>
    <script>
    console.log('here');
    require.config({
        baseUrl: 'foo',
    });
    </script>
    <script src="build/bundle.js"></script>
```

webpack:
```
  resolve: {
    alias: {
      'requirejs': 'requirejs/require'
    }
  }
```

console output:
```
> requirejs.s.contexts._.config.baseUrl
"foo/"
```

jdfreder:
> OK, externals is what I needed to use rather than resolve.alias, since we load our own require.js.

Good to know, thanks for keeping us posted.

PR #43, opened 2016-01-30
[Issue 15] Remove multer and use busboy for upload.

dalogsdon:
Fixes some errors caused by multer in #15


dalogsdon:
Whoops, only updated integration tests, forgot to update unit tests.

dalogsdon:
Rebased and updated unit tests. The unit tests turned out to be quite complicated with the move to Busboy, so I pushed what I had written, just the basic success cases. The integration tests also cover all the API paths.

parente:
All tests pass. Got a little spittle during the integration test here:

```
{ [Error: EXDEV: cross-device link not permitted, rename './data' -> './data2']
  errno: -18,
  code: 'EXDEV',
  syscall: 'rename',
  path: './data',
  dest: './data2' }
{ [Error: EEXIST: file already exists, mkdir './data'] errno: -17, code: 'EEXIST', syscall: 'mkdir', path: './data' }
```

Is this test only output or something that's going to appear all the time in the logs?

dalogsdon:
Right, I tested locally using `npm run integration-test` where it does not show the error. I can remove that since `make integration-test` runs the tests in docker. You will just have to manually clean up the data dir if you happen to run locally.

parente:
If it's test only dribble, it's fine to leave it.

app/notebook-store.js
```diff
@@ -44,10 +52,127 @@ function get(nbpath) {
...
+}
+
+function _fileFilter (filename) {
+    // check for *.ipynb extension
+    return /\.ipynb$/.test(filename);
+}
+
+var _uploadLimits = {
+    fields: 0,
+    files: 2, // really a limit of 1, but set to 2 to allow easy error response
```
jhpedemonte: wha? needs more details in that comment.

app/notebook-store.js
```diff
@@ -44,10 +52,127 @@ function get(nbpath) {
...
+    var fileCount = 0;
+
+    // only allow 1 file and nothing else in the form
+    var busboy = new Busboy({
+        headers: req.headers,
+        limits: _uploadLimits
+    });
+
+    // handle file upload
+    var uploadMessage = 'Make sure to upload a single Jupyter Notebook file (*.ipynb).';
```
jhpedemonte: Move this out of function, so it isn't constantly redefined.

test/integration/upload-notebook-test.js
```diff
@@ -0,0 +1,171 @@
...
+    after(function(done) {
+        // delete test data and move original data back
+        rimraf(dataDir, function(err) {
+            if (err) { console.error(err); }
+            fs.rename(dataDir2, dataDir, function(err) {
+                if (err) { console.error(err); }
+                done();
+            });
+        });
+    });
```
jhpedemonte: Couple of things:

1. We run integration tests in its own Docker container, separate from that of the proxy server. So I don't think moving this folder is necessary. @parente, am I understanding Docker functionality correctly?
2. I would prefer to treat the server as a black box during the integration tests. In this case, I would just upload your test files to a unique name (maybe generate a UUID-type string for the name).

parente: >  @parente, am I understanding Docker functionality correctly?

Yes.

dalogsdon:
I removed the data dir handling and changed the upload names to the form `it_XXXXXXXX` where X is a number. Also updated per @jhpedemonte's comments.

PR #44, opened 2016-02-01
Send kernel shutdown when client websocket closes

### LINKED ISSUE #25: NodeJS app should send kernel shutdown when client websocket disconnects
Most likely fix, call the jupyter-js-services shutdown method for the disconnecting websocket instance here: https://github.com/jupyter-incubator/dashboards_nodejs_app/blob/5ace686d7efd80eb1f70f3f8fc658c5a72504bf8/routes/api.js#L154 and in other errors cases too.

jtyberg:
Fixes #25

### DIFF
--- a/routes/api.js
+++ b/routes/api.js
@@ -106,8 +106,8 @@ function setupWSProxy(_server) {
         var kernelIdMatched = kernelIdRe.exec(req.url);
         if (kernelIdMatched) {
             var kernelId = kernelIdMatched[1];
-            socket.on('close', function () {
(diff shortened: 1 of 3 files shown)


routes/api.js
```diff
@@ -17,88 +18,122 @@ var kgBaseUrl = config.get('KG_BASE_URL');
 
 var server = null;
 var sessions = {}; // TODO remove old sessions, somehow
```
parente: Can this be cleaned up on the close callback also?

jtyberg: Added removal of sessions from memory in separate commit.

Issue #50, opened 2016-02-02
Use a real websocket library for decoding, filtering, encoding traffic


parente:
Right now we're using a somewhat home grown recipe that handles text frames well, but doesn't do anything with ping or pong or any of the other WS protocol op codes. This is biting us at the moment because the client's pongs are not making it to the kernel gateway in response to its pings, and so the KG kills the websocket connections after 90 seconds thinking they're dead.

I'm going to put in a quick hack fix for the pong problem, but we're pretty certain that it's not a robust solution. We need to move to something more robust eventually.

See sections 5.4 and 5.5.x of the websocket RFC https://tools.ietf.org/html/rfc6455#page-36

parente:
Maybe https://www.npmjs.com/package/non-native-websocket? Has a WebsocketFrame object?

jhpedemonte:
I did investigate that at one point. I don't remember why we passed on it. But worth another review.

parente:
@poplav Maybe take a look at how WebFrame in the library above might replace the logic in `ws-utils.js`. Just do it in a sample or branch or whatever.

parente:
More recent link: https://github.com/theturtle32/WebSocket-Node

poplav:
Looked into [ws express support](https://github.com/websockets/ws#expressjs-example).  I was able to setup a WebSocketServer in setupWSProxy from the http server passed in and receive a message, but am not able to call ```proxy.ws(req, socket, head);``` because I don't think we have access to those lower level [private variables](https://github.com/websockets/ws/blob/master/lib/WebSocket.js#L522) when using the library.  We get access to those variables from the native nodejs [upgrade event listener](https://github.com/jupyter-incubator/dashboards_nodejs_app/blob/master/routes/api.js#L82).  I don't think the route of using one of these higher level libraries is going to work with regard to proxying.

parente:
Fixed using `websocket` lib from npm.

PR #51, opened 2016-02-02
Quick fix for missing WS pong responses

parente:
Definitely a bandaid, definitely not the real fix, definitely gets us over an initial hump of lost connections every 90s

Ref #50


jhpedemonte:
Fine (for now, muahahahaha).

PR #52, opened 2016-02-02
[issue: 49] Send exceptions, stderr to console

### LINKED ISSUE #49: Hide stderr / exception tracebacks in dashboard view
Port of the styling added to `jupyter-incubator/dashboards` and `jupyter-incubator/dashboards_bundlers` recently.

parente:
Fixes #49

### DIFF
--- a/public/js/dashboard.js
+++ b/public/js/dashboard.js
@@ -64,6 +64,7 @@ requirejs([
     function _consumeMessage(msg, outputArea) {
         var output = {};
         var content = msg.content;
+        outer:
         switch (msg.header.msg_type) {
             case 'clear_output':
               outputArea.clear(content.wait);
@@ -73,8 +74,9 @@ requirejs([
               output.text = content.text;
               switch(content.name) {
                   case "stderr":
-                    output.name = StreamName.StdErr;
-                    break;
+                    // show stderr in console, no in the dashboard itself
+                    console.error(content.name, content.text);
+                    break outer;
                   case "stdout":
                     output.name = StreamName.StdOut;
                     break;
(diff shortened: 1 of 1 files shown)


public/js/dashboard.js
```diff
@@ -64,6 +64,7 @@ requirejs([
     function _consumeMessage(msg, outputArea) {
         var output = {};
         var content = msg.content;
+        outer:
```
jhpedemonte: Is this a GOTO? I would prefer a better way to handle this, if possible.

parente: Me too. Can either return, or rewrite the whole thing without case statements.

parente:
Added one last fix where stderr was still getting dumped into the page. Should be ready to :shipit: .

public/js/dashboard.js
```diff
@@ -61,50 +61,61 @@ requirejs([
         $container.removeClass('invisible');
     }
 
+    // delegates for kernel message handling, keyed by message type
+    // all functions receive the entire response message and a reference to an
+    // outputArea model associated with the code / widget that made the initial
+    // request associated with the received response
```
jhpedemonte: Nit: This seems like multiple sentences. Add caps and punctuation to make it more readable.

public/js/dashboard.js
```diff
@@ -61,50 +61,61 @@ requirejs([
...
+            output.outputType = OutputType.ExecuteResult;
+            output.data = msg.content.data;
+            output.metadata = msg.content.metadata;
+            output.execution_count = msg.content.execution_count;
+            outputArea.add(output);
+        },
+        error: function(msg, outputArea) {
+            // show tracebacks in the console, not on the page
+            var traceback = msg.content.traceback.join('\n');
+            console.error(msg.content.ename, msg.content.evalue, traceback);
```
jhpedemonte: In the future, will we want to show something in the page, so user knows that a given section of dashboard failed? Otherwise, user may not know to look in console for error details. Maybe a simple red icon?

jhpedemonte:
Didn't test, but looks good. Ready to merge after comment update.

public/js/dashboard.js
parente: Probably in the case of error specifically. stderr dribbles warnings all over the place and really isn't a good indicator of a true problem.

I can open that as a follow on enhancement. Some kind of error doohicky in the nav bar. For the moment, let's just replicate what we had using Thebe: sending errors to the console.

parente: Opened as #59

PR #54, opened 2016-02-03
[Issue 37] - Suppress stack trace on 404 errors.

poplav:
(c) Copyright IBM Corp. 2016

### DIFF
--- a/app.js
+++ b/app.js
@@ -103,7 +103,7 @@ app.use('/api', apiRoutes);
 
 /// catch 404 and forward to error handler
 app.use(function(req, res, next) {
-    var err = new Error('Not Found');
+    var err = new Error('404 Not Found: ' + req.url);
     err.status = 404;
     next(err);
 });
@@ -114,13 +114,15 @@ app.use(function(err, req, res, next) {
     if (app.get('env') === 'development') {
         // send stacktrace in development mode
         stacktrace = err.stack;
-        console.log("STACK:",err.stack);
+        if (err.status >= 500 && err.status < 600) {
+            console.log("STACK:",err.stack);
+        }
     }
 
     var status = err.status || 500;
     res.status(status);
 
-    // default to json, only send html if explicity requested
+    // default to json, only send html if explicitly requested
     if (req.accepts('html') &&
         !(req.accepts().length === 1 && req.accepts('*/*'))) {
         res.render('error', {
--- a/app.js
+++ b/app.js
@@ -101,26 +101,25 @@ else if(seedUsername && seedUsername !== "" && seedPassword && seedPassword !== 
 app.use('/', routes);
 app.use('/api', apiRoutes);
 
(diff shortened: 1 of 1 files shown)


jtyberg:
I don't think you need all of this.  Just remove the `next()` call from the 404 handler and the stack trace handler will never be called.

app.js
```diff
@@ -112,15 +112,18 @@ app.use(function(req, res, next) {
 app.use(function(err, req, res, next) {
     var stacktrace = '';
     if (app.get('env') === 'development') {
-        // send stacktrace in development mode
         stacktrace = err.stack;
-        console.log("STACK:",err.stack);
+        console.error("ERROR:",err.message);
+        if(err.status !== 404) {
```
dalogsdon: I think the stack trace could only be shown for all 500-level errors. Perhaps here we could do:
```javascript
if (err.status >= 500 && err.status < 600)
```
All 400-level messages will be explicitly handled and should therefore be descriptive enough to not require a stack trace.

poplav:
Removed handling the 404 in the next handler.  Also will only print stack traces in the 500 range.

app.js
```diff
@@ -114,13 +114,15 @@ app.use(function(err, req, res, next) {
     if (app.get('env') === 'development') {
         // send stacktrace in development mode
         stacktrace = err.stack;
-        console.log("STACK:",err.stack);
+        if (err.status >= 500 && err.status < 600) {
```
jtyberg: Looks like we're logging the stack trace if between 500 and 600, then we check the status code later, and if not set, we set it to 500 and we don't log the stack trace?

I propose explicitly handling the 404's by calling `res.send` and not calling `next`.  The catch-all handler should always print a stack trace, because it's a last resort, in a "we have no idea how we got here" kinda way.

app.js
```diff
@@ -101,11 +101,12 @@ else if(seedUsername && seedUsername !== "" && seedPassword && seedPassword !==
...
 
-/// catch 404 and forward to error handler
+/// catch 404 errors
 app.use(function(req, res, next) {
-    var err = new Error('Not Found');
+    var err = new Error('404 Not Found: ' + req.url);
     err.status = 404;
-    next(err);
+    res.status(err.status);
+    res.send(err);
```
dalogsdon: Content-type?

app.js
```diff
@@ -103,7 +103,7 @@ app.use('/api', apiRoutes);
 
 /// catch 404 and forward to error handler
 app.use(function(req, res, next) {
-    var err = new Error('Not Found');
+    var err = new Error('404 Not Found: ' + req.url);
```
dalogsdon: This is going to cause "404" to show twice in the error page.

dalogsdon: Also, why do we need to show the URL? It is going to be obvious what the URL was.

PR #60, opened 2016-02-04
[Issue: 53] Widget display data fixes

### LINKED ISSUE #53: Get a demo with widgets and plots working
Go beyond the basic simple / web3. See what happens.

parente:
Fixes #53

### DIFF
--- a/public/js/widget-manager.js
+++ b/public/js/widget-manager.js
@@ -36,10 +36,10 @@ define(['jupyter-js-widgets'], function(Widgets) {
     WidgetManager.prototype = Object.create(Widgets.ManagerBase.prototype);
 
     /*
-     * Called when a ipywidget is added to the DOM.
+     * Called when a jupyter widget is added to the DOM.
      *
      * msg: iopub channel, comm_msg with display data
-     * view: ipywidget view instance
+     * view: jupyter widget view instance
      * options: object with metadata to be associated with the view
      */
     WidgetManager.prototype.display_view = function(msg, view, options) {
@@ -61,7 +61,7 @@ define(['jupyter-js-widgets'], function(Widgets) {
     };
 
     /*
-     * Returns callbacks to be invoked in response to ipywidget triggered
+     * Returns callbacks to be invoked in response to jupyter widget-triggered
      * comm messages on shell, input, and iopub channels. Each callback takes
      * one parameter: the comm message from the kernel. The return value should
      * be an object with any of the following properties defined as functions.
@@ -72,7 +72,7 @@ define(['jupyter-js-widgets'], function(Widgets) {
      * shell.reply
      * input
      *
-     * view: ipywidget view instance
+     * view: jupyter widget view instance
      */
     WidgetManager.prototype.callbacks = function(view) {
         var options = view.options;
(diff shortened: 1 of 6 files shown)


dalogsdon:
Do you want to push the cell outputs?

parente:
This isn't working yet at all. I'm just putting a placeholder here so it's easier to look at the code together. I don't have a solution for how to get the cells associated yet.

public/js/widget-manager.js
```diff
@@ -45,16 +61,73 @@ define(['jupyter-js-widgets'], function(Widgets) {
...
+     * iopub.status
+     * iopub.clear_output
+     * iopub.output
+     * shell.reply
+     * input
+     *
+     * view: ipywidget view instance
+     */
+    WidgetManager.prototype.callbacks = function(view) {
+        // TODO: only registering one callback until https://github.com/ipython/ipywidgets/pull/353
```
parente: This needs to change once the bug is fixed in ipywidgets. But the callback API might change in the shim too.

Since we pin to a git sha for ipywidgets today, this "fix" will continue to work until we move on.

public/js/widget-manager.js
```diff
@@ -45,16 +61,73 @@ define(['jupyter-js-widgets'], function(Widgets) {
...
+        // is fixed
+        var mgr = this;
+        var callbacks = {
+            iopub : {
+                output: function(msg) {
+                    console.debug(msg.msg_type, view);
+                    // TODO: it's probably not always parent
+                    mgr.msgHandler(msg, view.options.parent.options.outputAreaModel);
```
parente: Look up the ancestor chain until we find the view.options that has `outputAreaModel` or the parent is null or `options.root: true` (which should be the same as `parent = null`).

public/js/widget-manager.js
```diff
@@ -34,10 +35,24 @@ define(['jupyter-js-widgets'], function(Widgets) {
     };
     WidgetManager.prototype = Object.create(Widgets.ManagerBase.prototype);
 
+    /*
+     * Called when a ipywidget is added to the DOM.
```
jhpedemonte: Are we going away from the "ipywidget" name? Maybe "Jupyter widget"?

public/js/widget-manager.js
```diff
@@ -45,16 +60,83 @@ define(['jupyter-js-widgets'], function(Widgets) {
...
+        }
+
+        var callbacks = {};
+        if(options.outputAreaModel) {
+            var mgr = this;
+            // TODO: only registering one callback until https://github.com/ipython/ipywidgets/pull/353
+            // is fixed
+            callbacks.iopub = {
+                output: function(msg) {
+                    mgr.msgHandler(msg, view.options.parent.options.outputAreaModel);
```
jhpedemonte: This should be `options.outputAreaModel`, right? After `while` loop above?

parente: Duh. Yes.

public/js/widget-manager.js
parente: Yes. That's true. The JS side is jupyter widget. The backend side is still an ipywidget (for now).

jhpedemonte:
Looks good. Will merge soon.

PR #66, opened 2016-02-05
Delete kernels over http and https protocols

jtyberg:
We are deleting kernels by invoking the kernel gateway API, but the code was using the `http` module, which does not work if kernel gateway is running over https.   

Switch to requests, which handles both http and https protocols.


parente:
Did you add request to the `package.json`?

jtyberg:
It's already there: 

https://github.com/jupyter-incubator/dashboards_nodejs_app/blob/master/package.json#L29

PR #72, opened 2016-02-08
[Issue 55] Add media queries to adjust dashboard width as the notebook does.

### LINKED ISSUE #55: Max dashboard layout fixed width
simple_dash

User lays out dashboard in the (mostly) fixed-width environment of the notebook server. We should match that experience when we display the notebook here. Depending on the browser width, the dashboard container width should snap to some maximum. (See the notebook CSS or the dashboards_bundlers CSS for what those breakpoints are.)

dalogsdon:
fixes #55

### DIFF
--- a/less/dashboard.less
+++ b/less/dashboard.less
@@ -57,6 +57,15 @@ h1 {
              */
         }
 
+        @media (min-width: 768px) {
+            width: 768px;
+        }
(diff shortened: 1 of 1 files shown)


parente:
Tested with the dashboards_in_a_nutshell notebook. Looks consistent with the notebook.

PR #75, opened 2016-02-08
[Issue 69] Read default dashboard parameters from config.

### LINKED ISSUE #69: Read Urth.* defaults from config
Currently hardcoded in dashboard.handlebars. See TODO at the bottom of that file.

dalogsdon:
Adds `config` handlebars helper and uses it to fill in default dashboard parameters. 

I went ahead and made use of this to remove the pervasive `authEnabled` flag being used on many template renders.

I also renamed the global `Urth` object to `Config` in the dashboard page since I was editing that template. This looked like the only use of "urth" outside of reading the notebook files.

Fixes #69

### DIFF
--- a/app/handlebars-helpers.js
+++ b/app/handlebars-helpers.js
@@ -2,12 +2,17 @@
  * Copyright (c) Jupyter Development Team.
  * Distributed under the terms of the Modified BSD License.
  */
-commonmark = require('commonmark');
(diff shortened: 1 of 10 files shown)


parente:
Tests pass. Manually tried a bunch of the notebooks with and without auth. No surprises.

PR #76, opened 2016-02-09
[Issue 74] - Filter hidden files from the notebook list.

poplav:
(c) Copyright IBM Corp. 2016

### DIFF
--- a/app/notebook-store.js
+++ b/app/notebook-store.js
@@ -48,6 +48,7 @@ function list() {
             if (err) {
                 reject(new Error('Error getting list of notebooks'));
             } else {
+                files = files.filter(_hiddenFileFilter);
                 resolve(files);
             }
         });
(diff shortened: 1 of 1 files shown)


app/notebook-store.js
```diff
@@ -92,6 +93,11 @@ function _getFilename (req) {
     return name;
 }
 
+function _hiddenFileFilter (filename) {
+    //check for hidden files or directories, returns true if file is not hidden
+    return !(/(^|\/)\.[^\/\.]/g).test(filename);
```
dalogsdon: Since we are simply doing a `readdir` to get a list of directory contents, can we just check for leading dot?

```javascript
return /^[^.]/.test(filename)
```

poplav: Yes, we could since we can assume readdir will return the content only.  The above might be more complete, but not not needed.  I'll update the PR.

dalogsdon: Yeah, I'd do the minimum for simplicity. A more complex solution can be used in the future if the need arises.

dalogsdon:
Works fine, tested with hidden files and folders.

Issue #85, opened 2016-02-10
Multiple linked widgets do not render, error in JS console


parente:
Bottom left panel in https://github.com/jupyter-incubator/dashboards/blob/master/etc/notebooks/test/test_layout_basic.ipynb should have 3 linked widgets. Only one of them seems to render.  The other two generate "Could not create view" errors in the JS console

```
Error: Could not create view↵    at new Error (native)↵    at Error.WrappedError (http://127.0.0.1:3000/components/jupyter-js-widgets.js:31415:26)↵    at promiseRejection (http://127.0.0.1:3000/components/jupyter-js-widgets.js:31493:34
```

<img width="1147" alt="screen shot 2016-02-09 at 11 09 16 pm" src="https://cloud.githubusercontent.com/assets/153745/12938497/87162978-cf82-11e5-8810-4984397229c0.png">

parente:
The `delete this._pendingExecutions[msg.parent_header.msg_id];` line is causing the widgets to go missing. Removing it allows additional widgets associated with the execute to appear. But when they do, they're not v-stacked properly like in the notebook. Probably styling at this point.

<img width="529" alt="screen shot 2016-02-09 at 11 23 08 pm" src="https://cloud.githubusercontent.com/assets/153745/12938619/3b492700-cf84-11e5-886c-88d42a9d8eb2.png">

jhpedemonte:
Closed by #86.

PR #86, opened 2016-02-10
[issue: 85] Render multiple widgets

parente:
Partial fix. Still fails to stack widgets vertically.

Ref #85

### DIFF
--- a/public/js/widget-manager.js
+++ b/public/js/widget-manager.js
@@ -44,8 +44,9 @@ define(['jupyter-js-widgets'], function(Widgets) {
      */
     WidgetManager.prototype.display_view = function(msg, view, options) {
         var widgetInfo = this._pendingExecutions[msg.parent_header.msg_id];
-        // no longer need to track
-        delete this._pendingExecutions[msg.parent_header.msg_id];
+
+        // keep tracking the message ID in case more than one widget will reside
+        // in the widget area
 
         view.options = view.options || {};
         view.options.outputAreaModel = widgetInfo.outputAreaModel;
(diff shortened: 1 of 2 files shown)


parente:
Fixed the layout issue. jupyter-js-widgets (ipywidgets) expects a `widget-subarea` within a `widget-area` in its CSS rules. I matched that DOM structure and now stacked widgets render properly.

I went back and manually tested all the other notebooks in `./data` to confirm the new DOM structure doesn't throw anything else off. Everything else appears to be working normally.

screen shot 2016-02-11 at 8 50 54 am

@jhpedemonte can you review and merge? Hopefully this doesn't cause too much breakage with shims you put in for decl widgets.

public/js/widget-manager.js
```diff
@@ -44,8 +44,9 @@ define(['jupyter-js-widgets'], function(Widgets) {
      */
     WidgetManager.prototype.display_view = function(msg, view, options) {
         var widgetInfo = this._pendingExecutions[msg.parent_header.msg_id];
-        // no longer need to track
-        delete this._pendingExecutions[msg.parent_header.msg_id];
+
+        // keep tracking the message ID in case more than one widget will reside
+        // in the widget area
```
jhpedemonte: So when do we clean this up?

parente: Apparently we can't, ever. There's no indication when some message might come from the kernel instantiating a new widget. Could be delayed and out of order with other messages. Could be dynamic based on the notebook logic.

I was probably in the wrong to try to prematurely optimize here. Since the nodes never go away in the DOM (yet) anyway, there's really no harm in keeping them around. And it sure looks like we have to anyway.

jhpedemonte:
Looks good. Merging.

PR #89, opened 2016-02-10
[Issue 84] Use some rendered_html styles from notebook.

### LINKED ISSUE #84: Heading and font size stylings differ between notebook dash preview and dashboard server
Use https://github.com/jupyter-incubator/dashboards/blob/master/etc/notebooks/test/test_layout_basic.ipynb  as a reference. In dashboard server, the h1 is smaller than the h2, for example. Headings not bolded the same way. And so on. Makes layout difficult to get right.

screen shot 2016-02-09 at 11 09 35 pm

vs 

screen shot 2016-02-09 at 11 09 16 pm

dalogsdon:
This will fix many cases involving markdown cells. 
This should also fix text getting cut off on the bottom in certain cases.

Resolves #84

### DIFF
--- a/app/handlebars-helpers.js
+++ b/app/handlebars-helpers.js
@@ -14,7 +14,7 @@ module.exports = {
     },
 
     mapCellType: function(cellType) {
-        return cellType === 'markdown' ? 'text-cell' : 'code-cell';
+        return cellType === 'markdown' ? 'text-cell rendered_html' : 'code-cell';
     },
 
     markdownContent: function(cellType, source) {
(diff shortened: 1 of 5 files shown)


parente:
Tested the new notebook and other demo notebooks. Improvement for sure. Did find one mismatch between the notebook dashboard view and here when the layout flips to the skinniest width. I think the behavior here is better, however.

Dashboard server:

screen shot 2016-02-11 at 8 14 27 am

Notebook server:

screen shot 2016-02-11 at 8 14 30 am

Think that should be opened against dashboards extension?

PR #91, opened 2016-02-11
[Issue 83] Fix look of `test_widget_overlap` dashboard.

### LINKED ISSUE #83: Table styles missing
Try https://github.com/jupyter-incubator/dashboards/blob/master/etc/notebooks/test/test_widget_overlap.ipynb in the notebook  dashboard preview and then here in the dashboard server. The table styling from the notebook environment is missing here. 

Related? When the table is wider than the window width, it does not overflow or scroll. In the notebook server, it overflows outside the allocated grid cell. The behavior should be consistent with the notebook.

screen shot 2016-02-09 at 11 06 55 pm

vs

screen shot 2016-02-09 at 11 05 29 pm

dalogsdon:
Adds table styling and introduces bootstrap CSS compilation.

Resolves #83

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,6 +14,7 @@
   },
   "dependencies": {
     "body-parser": "^1.13.3",
+    "bootstrap": "3.3.6",
     "busboy": "^0.2.12",
     "cookie-parser": "^1.3.3",
     "es6-promise": "^3.0.2",
--- a/views/dashboard.handlebars
+++ b/views/dashboard.handlebars
@@ -4,7 +4,6 @@
 }}
 
 {{! Required for jupyter-js-widgets }}
-<link rel='stylesheet' href='/components/bootstrap/css/bootstrap.min.css'>
 <link rel='stylesheet' href='/components/jquery-ui/jquery-ui.min.css'>
 <link rel='stylesheet' href='/components/widgets.min.css'>
 {{! Required for gridstack }}
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -89,10 +89,7 @@ gulp.task('copy:components', function() {
     var c3 = gulp.src([
             './bower_components/jquery-ui/themes/smoothness/images/**/*'
         ]).pipe(gulp.dest('./public/components/jquery-ui/images'));
-    var c4 = gulp.src([
-            './node_modules/bootstrap/dist/**/*'
-        ]).pipe(gulp.dest('./public/components/bootstrap'));
-    return merge(c1, c2, c3, c4);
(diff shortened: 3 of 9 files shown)


package.json
```diff
@@ -14,6 +14,7 @@
   },
   "dependencies": {
     "body-parser": "^1.13.3",
+    "bootstrap": "3.3.6",
```
jhpedemonte: This is fixed to this version. Do we want to instead accept any 3.x changes?  i.e. `"^3.3.6"`?  That would match what we do elsewhere, although we've also talked about changing to fixed versions (but more so for the jupyter libs)...

dalogsdon: Yes, I fixed it to match notebook's bootstrap version.

jhpedemonte:
Testing now. Noticed that in our server, the table does not overflow as it does in the Notebook. However, the selector that implements the overflow: hidden rule comes from Phosphor. I don't think we should change this -- I would assume that the Notebook will change to match our server once it goes "full Phosphor".

views/dashboard.handlebars
```diff
@@ -4,7 +4,6 @@
 }}
 
 {{! Required for jupyter-js-widgets }}
-<link rel='stylesheet' href='/components/bootstrap/css/bootstrap.min.css'>
```
jhpedemonte: Since we are no longer using this, you can remove lines 92-94 in **gulpfile.js**.  With that change, feel free to merge.

parente:
>  I don't think we should change this -- I would assume that the Notebook will change to match our server once it goes "full Phosphor".

Agreed. We should be looking ahead wherever possible.

> Looks like the notebook adds it when the message type is "text/html", so we could wrap the outputAreaModel.add call with a message type check to selectively add the proper output classes.

Looks like the output area object already has an addClass method on the view. Maybe that's a better approach: https://github.com/jupyter/jupyter-js-output-area/blob/51e6158f1f98e74e4761473a506da98d7680dbeb/src/widget.ts#L87

dalogsdon:
I removed some other references to bootstrap CSS according to @jhpedemonte.

I used the `addClass()` function from `OutputAreaWidget` according to @parente.

I still need to apply the `rendered_html` class at the proper time instead of on all output types.

parente:
In the `dashboard.js#consumeMessage`, I wonder if you could add it "just in time" based on the message type and its content. That function currently receives a model, though.

parente:
Tested. Merging this with the global application of `.rendered_html` for now. Opened a follow on plumbing issue to clean that up later: #93.

Issue #77, opened 2016-02-09
Dashboard list shows sub directories.


dalogsdon:
The dashboard list page shows all contents of the data directory, which includes sub directories. However, selecting an entry in the list will attempt to load that as a notebook.

The `GET /dashboards/*` API does allow for sub directories, e.g. `GET /dashboards/subdir/foo.ipynb`. 

Should we support sub directories in the list page or remove support from the GET API?

parente:
At some point in the near future, we might end up posting more than a notebook, but also frontend assets like imported declarative widgets, stylesheets, user defined styles, etc. In those cases, subdirs will sure help make sure that dependencies don't wind up overwriting one another.

dalogsdon:
I have the initial implementation done in a branch. The tests don't pass, so I need to fix those.

Issue #95, opened 2016-02-15
No link to dashboard list if an index notebook exists


dalogsdon:
Should the user be able to navigate to the dashboard list (without editing URL manually) if the index notebook exists? Currently the "Home" link in the navbar links to the root URL, so that will show the index notebook. Should there be a separate link to the list (that is shown only if index notebook exists), or would the index notebook be expected to contain a link if desired?

parente:
Yea, we discussed this (having two links in the nav bar) and took no action to start to keep it simple. We can address it in this issue.

dalogsdon:
![image](https://cloud.githubusercontent.com/assets/6362426/13189384/ad434c60-d71c-11e5-801f-2b42763e3de0.png)

dalogsdon:
![image](https://cloud.githubusercontent.com/assets/6362426/13189414/dcebfde0-d71c-11e5-875a-3a3a2ce0135a.png)

parente:
Is the secondary link visible all the time or only when there's an index.ipynb?

dalogsdon:
only when you are at the url `/dashboards/index.ipynb`

parente:
Even better.

Issue #88, opened 2016-02-10
Style login/list pages


dalogsdon:
Currently we have default bootstrap style login and dashboard list. If we do issue #87, we could follow up with login and list styling to match. I put these in a separate issue as I feel they are less important than the navbar since they reside in their own pages which are not the main focus of dashboards.

parente:
Also add better titles to the HTML. List page has none. Check login.

dalogsdon:
List page got styled in conjunction with #87.

PR #96, opened 2016-02-15
[Issue 77] List dirs and files

dalogsdon:
Should work the same as in notebook UI. Use of `/dashboards` will function properly with directories.


dalogsdon:
Resolves #77

parente:
Integration tests pass. Looks reasonable enough to me. We'll tackle how dashboard assets like declarative widget dependencies are shoved into folders per notebook in #15 . We should also improve the navigation back to the list in #95 or here. If i'm a few folders deep, should the "Jupyter Dashboards" link take me back to the root of the tree or to the view of the current folder. (In Jupyter Notebook, it goes to the current folder.)

I don't have any comments on the code itself.

dalogsdon:
FWIW, I did end up adding a breadcrumb navigation (like in the notebook) to the list page as part of #88. This can help navigation, but doesn't solve #95.

PR #105, opened 2016-02-17
[Issue 67] Use jupyter-js-notebook for output area.

dalogsdon:
Resolves #67

PR #106, opened 2016-02-18
[issue: 82] %%javascript does not execute properly

nitind:
Update from now-deprecated jupyter-js-output-area to jupyter-js-notebook,
  document the Makefile's clean target, and have the default config.json
  set to listen on all interfaces.
(c) Copyright IBM Corp. 2016

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,7 @@ help:
 	@echo '   logging-logging - demo + node network logging enabled'
 	@echo '              test - run unit tests'
 	@echo '  integration-test - run integration tests'
+	@echo '             clean - removes built Docker images, bower components, node modules, and certificates'
 	@echo
 	@echo 'Dashboard server option defaults (via nconf):'
 	@cat config.json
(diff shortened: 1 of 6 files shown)


nitind:
Conflicts with PR#105.

Makefile
```diff
@@ -38,6 +39,7 @@ clean:
 	@-rm -rf public/components
 	@-rm -rf public/css
 	@-rm -rf public/urth_components
+	@-docker rmi -f $(KG_IMAGE_NAME) $(DASHBOARD_IMAGE_NAME)
```
parente: Probably don't want to clean up the docker image every time we clean. The caching should be smart enough to rebuild what's needed. Is that not working for you?

config.json
```diff
@@ -7,7 +7,7 @@
     // Mode in which the dashboard server express environment runs
     NODE_ENV: development
     // Interface on which the dashboard server listens
-    IP: 127.0.0.1
+    IP: 0.0.0.0
```
parente: Definitely want to keep 127.0.0.1 as the default so the server comes up on localhost only unless the user purposely overrides it (like we do in the Dockerfile.server).

parente:
I merged #105 because it fixes the original issue without minimal changes. Let me know if the `make clean` part of this should still get merged.

Makefile
nitind: It is, it just seemed incongruous that it builds the images but never removes them.

PR #115, opened 2016-02-19
[Issue 59] Error indicator in navbar

### LINKED ISSUE #59: Indicate a problem when the kernel sends an error message
Strawman: some indicator in the nav bar at the top. Sort of like what notebook does when a kernel dies or a websocket disconnects or ... Don't hard code the indicator to kernel errors only though: it's very likely we're going to want to use that same indicator area to give other types of status too.

dalogsdon:
Resolves #59

### DIFF
--- a/public/js/dashboard.js
+++ b/public/js/dashboard.js
@@ -24,11 +24,13 @@ requirejs([
     'jquery',
     'gridstack-custom',
     'jupyter-js-output-area',
+    'jupyter-js-services',
     'jupyter-js-widgets',
     'urth-widgets',
     'widget-manager',
+    './error-indicator',
     './kernel'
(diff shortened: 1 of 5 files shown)


public/js/dashboard.js
```diff
@@ -137,6 +149,7 @@ requirejs([
     Kernel.start().then(function(kernel) {
         // initialize an ipywidgets manager
         var widgetManager = new WidgetManager(kernel, _consumeMessage);
+        _handleError(kernel);
```
parente: Nit: name suggests to me this function is the callback. But it's really registering the callback. Maybe `_initErrorHandler`?

dalogsdon: Sure, I'll update

parente:
Tested locally. Confirmed working with the lorenz example which has an innocuous NoneType error near the bottom of the notebook.

PR #116, opened 2016-02-19
[Issue 95] Show a link to dashboard list when viewing index.ipynb

### LINKED ISSUE #95: No link to dashboard list if an index notebook exists

dalogsdon:
Resolves #95, depends on PR #59

### DIFF
--- a/less/common.less
+++ b/less/common.less
@@ -15,6 +15,17 @@
         }
     }
 
+    .error-container {
+        position: relative;
+
+        .error-indicator {
+            color: @error-red;
+            cursor: pointer;
+            height: @navbar-height;
+            padding: 13px;
+        }
+    }
+
     .logout-btn,
     .login-btn {
         padding-left: 0;
@@ -26,7 +37,6 @@
         padding: 0;
         border: none;
     }
-
     .navbar-nav {
         margin: 0;
     }
--- a/less/common.less
+++ b/less/common.less
@@ -16,8 +16,6 @@
     }
 
     .error-container {
-        position: relative;
-
         .error-indicator {
             color: @error-red;
             cursor: pointer;
--- a/less/common.less
+++ b/less/common.less
@@ -3,7 +3,7 @@
  * Distributed under the terms of the Modified BSD License.
  */
 
-// Simplified navbar
+// Customize navbar
 .navbar {
     border-top: none;
     border-right: none;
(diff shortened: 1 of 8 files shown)


parente:
Works as advertised. Tested by putting an index.ipynb in the root and seeing the link. Will wait for #59 before merging.

I think we need to pause on adding more navigation options now. It's not clear to me yet if the common use case is going to be to have a dashboard server with multiple dashboards where the chrome and nav are heavily used, or if the more common case is deploying a chromeless dashboard server with exactly one dashboard notebook in it.

parente:
When I try to run the lorenz demo against this PR:

```
SyntaxError invalid syntax (<ipython-input-1-6ca80ea561f3>, line 1) [1;36m  File [1;32m"<ipython-input-1-6ca80ea561f3>"[1;36m, line [1;32m1[0m
[1;33m    In this Notebook we explore the Lorenz system of differential equations:[0m
[1;37m          ^[0m
[1;31mSyntaxError[0m[1;31m:[0m invalid syntax
```

I left my browser as-is, switched back to master, and try it there. Bug goes away. So there's something in this PR screwing with the cell indexing such that the markdown in the notebook is being sent to the kernel for execution instead of the proper code.

Also, I tried overlaying #117 on top of this PR in a local branch, but I get the same problem. So this is something new.

parente:
More info. With this PR, all DOM nodes are getting marked as `code-cell`:

screen shot 2016-02-22 at 1 16 39 pm

In master, they're not:

screen shot 2016-02-22 at 1 10 52 pm

dalogsdon:
Looks like there was an update to handlebars paths (https://github.com/wycats/handlebars.js/issues/1028) so the path we used for hidden cells was not valid. Should be resolved now.

parente:
Pulled, make dev-install, make dev. Get another exception in lorenz notebook.

```
SyntaxError invalid syntax (<ipython-input-1-25e95ea467e7>, line 1) [1;36m  File [1;32m"<ipython-input-1-25e95ea467e7>"[1;36m, line [1;32m1[0m
[1;33m    First, we import the needed things from IPython, NumPy, Matplotlib and SciPy.[0m
[1;37m                   ^[0m
[1;31mSyntaxError[0m[1;31m:[0m invalid syntax
```

parente:
We paired. The latter exception is fixed by #117. Gridstack rearranges the DOM if the window is skinny on startup.

PR #117, opened 2016-02-19
[Issue 56] Ensure notebook order when executing cells.

### LINKED ISSUE #56: Simple demo sometimes says x is not defined
Like the message did not reach the kernel in order. Or at all.

dalogsdon:
We were using DOM order. Added data-cell-index attr to sort cells in notebook order.

Resolves #56


parente:
Tried this locally and everything seemed kosher. I'm holding off on merging it until the mystery in #116 is resolved.

PR #128, opened 2016-03-02
[issue: 127] Fix upload authorization header format

### LINKED ISSUE #127: Upload auth token format is inconsistent
Kernel gateway and other Jupyter projects adopt the GitHub auth token format:

```
Authorization: token <token value>
```

We're supporting this when we send a token to a kernel gateway. We're not supporting it in our own upload handler:

https://github.com/jupyter-incubator/dashboards_server/blob/b5f8c49ea7128526eea3f76a1fa50cc2db214adb/app/auth-token.js#L14

We should change to match. When we do, we also have to update the bundler. (Send both PRs off this one issue. Don't open another one in bundlers.)

https://github.com/jupyter-incubator/dashboards_bundlers/blob/master/dashboards_bundlers/server_upload/__init__.py#L43

parente:
`Authorization: token <value>` like Github

Fixes #127

### DIFF
--- a/app/auth-token.js
+++ b/app/auth-token.js
@@ -11,7 +11,7 @@ var authToken = config.get('AUTH_TOKEN');
 var authMessage = 'Authentication required. Please provide a valid auth token.';
 
 module.exports = function(req, res, next) {
-    if (authToken && authToken !== req.headers.authorization) {
+    if (authToken && ('token ' + authToken) !== req.headers.authorization) {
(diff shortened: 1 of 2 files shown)


parente:
Should merge in conjunction with https://github.com/jupyter-incubator/dashboards_bundlers/pull/22

parente:
Thanks @minrk!

PR #138, opened 2016-03-10
[issue: 123] Enable passport local auth strategy

### LINKED ISSUE #123: Pluggable authentication
We implemented a simple shared auth scheme for demonstration purposes. We need to consider how to make the auth pluggable. http://passportjs.org/ has served me well in the past.

We should:

* Support auth flows that have the user login with OAuth providers
* Continue to support auth flows where the user logs in using the login page the dashboard server provides

TODOs

* [x] enable passport
* [x] enable passport-local
* [x] make auth strategy configurable
* [x] fix flash messages on login failure
* [x] try configuring an oauth strategy (passport-twitter, passport-box)
* [x] integration tests for auth-local
* [x] landing page when not authenticated
* [x] update readme / wiki with info on how to configure auth

parente:
Fixes #123


parente:
Yea. Passport cuts down on a lot of custom code. I put a todo list over in the issue about what I have left to do. The proof will be in how easy/hard it is to swap in an oauth provider. The trickiness is in the fact that we don't need to expose our own login page at that point, so the routes need to be part of the pluggable local auth module and something probably needs to be the landing page in front of the auth redirect. Maybe a degenerate login page that just has a button saying "Click to login using XXXXX" with the name of the auth provider?

parente:
@jhpedemonte @dalogsdon I think this is ready for final review. There's a screenshot / animation in the associated issue #123. The wiki page with usage information is here https://github.com/jupyter-incubator/dashboards_server/wiki/Authentication

README.md
```diff
@@ -19,6 +19,7 @@ It is also has close ties to [jupyter-incubator/declarativewidgets](https://gith
 * Ability to run a Jupyter Notebook with [layout metadata](https://github.com/jupyter-incubator/dashboards) as a standalone dashboard application
 * Ability to navigate a list of multiple notebooks and select one to run as a dashboards
 * Optional shared login to secure access to the dashboard server
+* Ability to add custom authentication mechanisms usign the [Passport](http://passportjs.org/) middleware for Node.js
```
jhpedemonte: /usign/using/

app.js
```diff
@@ -75,24 +77,37 @@ app.use(session({
...
-            next();
-        }
-        else {
-            //save the previous page in the session to know where to redirect back to after login
-            req.session.redirectAfterLogin = req.path;
-            res.redirect('/login');
-        }
+    // Load auth strategy based on config
+    var strategy = require(config.get('AUTH_STRATEGY'))(app);
+    passport.use(strategy);
```
jhpedemonte: Does this fail correctly on startup if AUTH_STRATEGY isn't properly specified?

app/auth-box.js
```diff
@@ -0,0 +1,52 @@
...
+        // Don't bother if the user is already authenticated.
+        if(req.user) {
+            return res.redirect('/');
+        }
+        // Render the simple view included in the server. Substitute a custom
+        // view here as needed.
+        res.render('login', {
+            title: 'Log in',
+            provider_id: 'box', // when the user presses the login button goes to /login/{{provider_id}}
+            provider_name: 'box' // name to show on the button
```
jhpedemonte: /box/Box/

app.js
parente: A non-existent module crashes the app immediately.

`make dev AUTH_STRATEGY=./app/qwerqwer`

A module that doesn't return a strategy crashes the app immediately.

`make dev AUTH_STRATEGY=./app/auth-bogus`

implemented as:

```
module.exports = function() {
    return {};
};
```

parente:
Addressed the two typos. I think the bogus strategy setting fails properly when set incorrectly.

jhpedemonte:
Code looks good. Tested with local auth strategy and working as expected.

PR #142, opened 2016-03-15
[Issue: 31] Allow dashboard clients to reconnect to running kernel

### LINKED ISSUE #31: Allow dashboard clients to reconnect to running dashboards within a time window
Improve upon the immediate cleanup logic implemented in #25 to give the client time to try to reestablish its websocket connections. 

Strawman: Set a timer in the nodejs app. If a client with the `x-jupyter-session-id` header returns before the timer fires, cancel the timer and continue forwarding messages to the associated kernel. If the timer goes off before a client reconnects, kill the kernel.

Of course, this also requires that the frontend JS attempt reconnections. It looks like jupyter-js-services already supports it, but we should check if it works in our use case.

nitind:
Delay the reaping of disconnected kernels so the client has a chance to reconnect.

Fixes #31

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -25,4 +25,7 @@ config.set('NOTEBOOKS_DIR', path.join(__dirname, '..', config.get('NOTEBOOKS_DIR
 config.set('DB_FILE_EXT', '.ipynb');
 config.set('DB_INDEX', 'index.ipynb');
 
+// how long to keep a kernel alive and reconnectable after its WS connection drops
+config.set('KG_KERNEL_RETENTIONTIME', 120000);
+
 module.exports = config;
--- a/routes/api.js
+++ b/routes/api.js
@@ -19,9 +19,11 @@ var urlToDashboard = require('./url-to-dashboard');
 var kgUrl = config.get('KERNEL_GATEWAY_URL');
 var kgAuthToken = config.get('KG_AUTH_TOKEN');
 var kgBaseUrl = config.get('KG_BASE_URL');
+var kgKernelRetentionTime = config.get('KG_KERNEL_RETENTIONTIME');
 
 var server = null;
 var sessions = {};
+var disconnectedKernels = {};
 var apiRe = new RegExp('^/api(/.*$)');
 var kernelIdRe = new RegExp('^.*/kernels/([^/]*)');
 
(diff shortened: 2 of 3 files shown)


routes/api.js
```diff
@@ -199,8 +216,10 @@ var killKernel = function(kernelId) {
         method: 'DELETE',
         headers: headers
     }, function(err, response, body) {
-        debug('PROXY: kill kernel response: ' +
-            response.statusCode + ' ' + response.statusMessage);
+        if (response) {
```
jhpedemonte: What happens if we don't have a response here? Is that an error? If so, probably should be logged at the very least. If there is an error, is there anything we should do to recover/retry?

test/base/routes/api-test.js
```diff
@@ -210,8 +213,10 @@ describe('routes: api', function() {
                 expect(options.headers.Authorization).to.equal('token ' + kgAuthToken);
             }
             spy.restore();
+            fakeClock.restore();
             done();
-        }, 0);
+        }, kgKernelRetentionTime);
+        fakeClock.tick(kgKernelRetentionTime + 1000);
     });
 });
```
jhpedemonte: Should we also add an integration test for reconnection?

routes/api.js
parente: Second to logging it. Not sure a retry makes sense. From the kernel gateway side, I believe the only way a DELETE for a kernel can happen is if the kernel doesn't exist.

test/base/routes/api-test.js
parente: @nitind and I discussed this. Are any of the integration tests creating a real websocket client and testing communication through to a kernel gateway? (I didn't see one.) If not, then I don't think we should do all that at the moment just to get this in. As it stands, this is even hard to test manually when the browser and dashboard server are colocated on the same host. They really need to be on separate hosts with a network failure between them to see jupyter-js-services execute its reconnect logic.

routes/api.js
nitind: So far I think that's only happened because the test was originally a spy. I've updated it to be a stub instead, and have a "response" return value.

app/config.js
```diff
@@ -25,4 +25,7 @@ config.set('NOTEBOOKS_DIR', path.join(__dirname, '..', config.get('NOTEBOOKS_DIR
 config.set('DB_FILE_EXT', '.ipynb');
 config.set('DB_INDEX', 'index.ipynb');
 
+// how long to keep a kernel alive and reconnectable after its WS connection drops
+config.set('KG_KERNEL_RETENTIONTIME', 120000);
```
parente: Should the default be in config.json instead of here?

parente: I addressed this in #144.

parente:
Added one more commit in #144 to address default timeout value and config location.

PR #151, opened 2016-03-21
[Issue 150] Fix caching of notebook data

### LINKED ISSUE #150: Changes to redeployed of bundled dashboard do not appear in UI on refresh
Looks like we're now caching notebooks after first load somewhere server-side. Need to read from disk every time to account for redeploys.

Behavior used to be correct.

jhpedemonte:
Ensure that we're always storing/getting using same path
from request.

Fixes #150

### DIFF
--- a/app/notebook-store.js
+++ b/app/notebook-store.js
@@ -17,7 +17,6 @@ var INDEX_NB_NAME = config.get('DB_INDEX');
 var ZIP_EXT = '.zip';
 
 var allowedUploadExts = [ DB_EXT, ZIP_EXT ];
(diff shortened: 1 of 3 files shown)


jhpedemonte:
@parente Try this.

parente:
Aye aye! Trying!

parente:
Nope. Same behavior as before. Redeploying with a markdown cell change shows the old markdown in the rendered dashboard. Both for bundled and unbundled dashboards.

parente:
Those sneaky regexes! Much better.

Running integration tests here. If all green, will merge.

PR #153, opened 2016-03-21
Update ipywidgets to 0.0.17

jhpedemonte:
Latest release of jupyter-js-widgets in 0.0.17, which is later than the commit hash we're currently using.


jhpedemonte:
Unfortunately, widgets aren't registered anymore, so there's nothing in `ManagerBase._model_types`. This results in failure to load widgets.

The [**web4** example](https://github.com/ipython/ipywidgets/tree/master/jupyter-js-widgets/examples/web4) in the `ipywidgets` repo passes in widget state (see [index.html](https://github.com/ipython/ipywidgets/blob/master/jupyter-js-widgets/examples/web4/index.html)).  This sets the `model_module` property to `juptyer-js-widgets`, which seems to be important. However, I'm not sure what we need to do in our code to make this work. Will have to ask devs.

parente:
> Unfortunately, widgets aren't registered anymore, so there's nothing in ManagerBase._model_types. This results in failure to load widgets.

@jdfreder @SylvainCorlay We're looking to jump to the latest jupyter-js-widgets code in prep for stable release to coincide with Notebook 4.2b1. We had things working here back on ipython/ipywidgets@851fd648195be1f42cab2560600c3b629bbe3330 but it looks like something changed in the manager since then. Maybe in support of Jupyter Lab?

Is there something we're supposed to do to register the widgets? Or is that a notebook author's responsibility now?

SylvainCorlay:
@parente @jhpedemonte we don't use the widget at all registry anymore. Instead, even the core widgets use the mode_module + model_name mechanism (and view_module + view_name).

Hence widget models must have the two `_model_module` and `_view_module` attributes.

If you don't want to use the whole nbextension file, you can use a requirejs map to shorten it.

parente:
@SylvainCorlay Thanks for the pointer. We can dig in. It's not clear to me yet what we as a canvas need to do to make the widgets available so that when code comes back from a kernel to instantiate a widget, it's available for use on the frontend. (Our goto reference has been the web3 example which doesn't look like it was updated.)

@jhpedemonte Diff from our current commit to master:

https://github.com/ipython/ipywidgets/compare/851fd648195be1f42cab2560600c3b629bbe3330...master

Most relevant commits:

* https://github.com/ipython/ipywidgets/commit/d7550cd821bea71539a10a6090a7ccc7578bfd4c Fixes to web1 and web2 demos
* https://github.com/ipython/ipywidgets/commit/59ffb4a3dc158981194b396685648ae47ee30240 stop using registry internally
* https://github.com/ipython/ipywidgets/commit/10dc8d2fef1122c4cb2972cb3680ce81235aa321 Use requirejs module in every widget

jhpedemonte:
Most of the examples pass in a state, which includes `model_module` property. But as an embedder which is starting up a notebook from scratch, we have no state. So how are widgets initialized correctly now in v0.0.17?

SylvainCorlay:
Backbone widget models have default values for all these attributes specified with the `defaults` property. Is this what you were looking for?

jhpedemonte:
Ugh, so today everything seems to be working as expected, using this patch.  Which is odd!  Yesterday, I did a `make clean` before testing and couldn't get ipywidgets working. Today, they're fine. @parente is going to confirm on his machine.

Only thing missing is that Declarative Widgets also need to be updated to new registration code.

parente:
Failure mystery solved:

`make kill clean build dev-install`

If you've got an old kernel gateway container lying around or haven't rebuilt it (`kill` and `build`) the widget JS can't talk to the widget Python (duh). Matching all that up, and it works like a charm.

parente:
Before merging this, please rebundle the demo notebooks with the corresponding fix that makes decl widgets work again.

jhpedemonte:
Fixing this the "right" way for declwidgets won't be so easy. But we can continue to use the older register APIs, since we're working against ipywidgets 5.x. I put those methods back and now declwidgets demos are working as expected.

parente:
Crossref: https://github.com/jupyter-incubator/declarativewidgets/issues/275

I'll test one more time before merging.

jhpedemonte:
Added code to work with updated DeclWidgets code (see jupyter-incubator/declarativewidgets#275). But safe to commit before, works with older code too.

parente:
I think this should go in 0.4.0. I've tried all the examples, plus deploying and redeploying dashboards using dashboard_bundlers with the declarative_widgets 0.5.0 pre-release.

screen shot 2016-03-24 at 10 34 31 am

However, it bombs using the last stable release of decl widgets 0.4.2  with an exception:

`Uncaught Error: Script error for "base/js/namespace", needed by: nbextensions/urth_widgets/js/widgets/DeclWidgetModel`

So some backward compat problem. Hopefully minor.

parente:
After some more searching, looks like 0.4.2 has other compatibility problems here. So we're going to be dependent on 0.5.0 of declarative widgets and up. Not a huge problem, since decl widgets was already busted over here anyway in the last tagged version due to bundling issues.

Issue #161, opened 2016-03-25
Redirect to favicon.ico after login


parente:
Simple fix is probably to use the favicon middleware so that the ico is exposed and passport doesn't think it needs to redirect there instead of the proper / page.

<img width="867" alt="screen shot 2016-03-25 at 9 10 44 am" src="https://cloud.githubusercontent.com/assets/153745/14044706/e3f46100-f269-11e5-9620-a988a9e093b9.png">

jhpedemonte:
Why does passport redirect to `/favicon` instead of `/`?

jhpedemonte:
Odd, I saw this with the docker-compose all-in-one, but not when running from `make dev`.

jhpedemonte:
Added PR that may fix this by serving a favicon. Some things I have found:

1. Browsers will cache the favicon differently than regular cache. To see change, you'll need to restart browser. Even then, if you change the icon, restarting and clearing cache will still show old. Not sure how to reset this cache.
2. The redirect may be tied in to this. I tested on Safari and saw the new icon (from PR). Then when I removed the serving of the favicon, restarted Safari and restarted the server with login set, I saw this error again. But not 100% sure it is fixed since it's very hard to test.

jhpedemonte:
Oh, well found way to do so on Chrome: `rm ~/Library/Application\ Support/Google/Chrome/Default/Favicons`

parente:
> Actually, only happens once per browser, no matter how server is started. Even after clearing cookies/local data, still doesn't go back to /favicon.ico.

Yes. It's odd.

jhpedemonte:
Should we cut a 0.4.1 release for this?

parente:
Your choice. It's just a git tag. Not sure this bug fix warrants one vs all the other bugs we have open.

PR #163, opened 2016-03-25
Remove unused dependencies

jhpedemonte:
Don't see these being used. Ran through test notebooks, as well as unit and integration tests. Everything seems to be working.

cc @parente @dalogsdon @nitind - Please review and make sure I'm not removing something I shouldn't.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,9 +41,7 @@
     "serve-favicon": "^2.3.0",
     "tmp": "0.0.28",
     "url-join": "0.0.1",
-    "websocket": "^1.0.22",
-    "ws": "^0.8.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -38,12 +38,9 @@
     "passport-local": "^1.0.0",
     "request": "^2.67.0",
     "requirejs": "^2.1.21",
-    "serve-favicon": "^2.3.0",
```
parente: Might need this one to fix #161 .

parente: serve-favicon is now used.

jhpedemonte:
Updated to remove removal of `serve-favicon`.

PR #165, opened 2016-03-28
Remove npm from kernel gateway image

parente:
No longer needed because we don't need/want to build the ipywidgets JS in this image, and only need the Python


jhpedemonte:
Tested. Everything looks good. Merging.

Issue #156, opened 2016-03-24
Browser loses websocket connection, can't reconnect in streaming-meetup demo


parente:
I deployed streaming-meetup demo from https://github.com/jupyter-incubator/dashboards/blob/master/etc/notebooks/stream_demo/meetup-streaming.ipynb to the dashboard server as a bundle. Good news: it started off working just fine (sans a little styling nit on the paper card for some reason):

![screen shot 2016-03-24 at 1 40 00 pm](https://cloud.githubusercontent.com/assets/153745/14025807/46f497c4-f1c6-11e5-924e-e8d5de415f6e.png)

Bad news: after a few seconds, it stopped working. Checking the console I saw an endless cadence of:

![screen shot 2016-03-24 at 1 43 09 pm](https://cloud.githubusercontent.com/assets/153745/14025844/6b93ce10-f1c6-11e5-834f-b6e547fb0f11.png)

I left the page and noticed that the kernel did indeed get cleaned up after our idle timeout. But when I come back to the dashboard page anew, the same error appears endlessly no matter how many times I refresh. It's like the Websocket part of the dashboard server just disappeared.

I checked the dashboard server logs, no errors that I can see. Also no problems loading the HTTP resources on refresh. Just the websocket proxy that has gone kaput.

parente:
Tried again and hit it again. Here's the combined log up to the approximate point of failure. 

https://gist.github.com/parente/02b2fe67897a4dacedd1

The Spark logging is noisy, but searching through, I don't see any message from the dashboard server that things have gone awry.

parente:
So even crazier, I left it running after the websocket failure. After 200 or so console log errors about the websocket failure (see above), it reconnected and has been running ever since without a problem (30+ minutes!)

jhpedemonte:
Testing using `make dev-logging` (so **not** running dashboard server in a container). Have yet to be able to reproduce the issue. Will keep testing.

parente:
I seem to have hit this probably again this AM, but this time on the `POST /api/kernels` call, not the websocket. The call would hang for some time before the dashboard server logged:

```
POST /api/kernels?1459777914810 - - ms - -
```

This looks like proxy upstream timeout.

parente:
[...]

PR #167, opened 2016-04-04
Configure /api proxy with SSL options

### LINKED ISSUE #166: /api/* endpoint does not configure SSL properly
We're configuring HTTPS for the express server properly. The /api proxy endpoint relies on a createProxyServer call which must also accept an `ssl` parameter bundle with the same key/cert as the main server.

parente:
Fixes #166

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -5,6 +5,7 @@
 var nconf = require('nconf');
 var hjson = require('hjson');
 var path = require('path');
+var fs = require('fs');
 
 var config = nconf.argv()
                   .env()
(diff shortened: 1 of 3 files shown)


parente:
@dalogsdon please give this a look see

app/config.js
```diff
@@ -26,4 +27,20 @@ config.set('NOTEBOOKS_DIR', path.join(__dirname, '..', config.get('NOTEBOOKS_DIR
...
+    }
+    if(!fs.existsSync(cert_file_location)) {
+        throw new Error('Invalid file path for HTTPS_CERT_FILE');
+    }
+    config.set('SSL_OPTIONS', {
+       key: fs.readFileSync(key_file_location),
+       cert: fs.readFileSync(cert_file_location)
+    });
+}
+
```
dalogsdon: :+1: for moving config validation to config.js

dalogsdon:
The change works for me, but I also see things work without the change. Looked into it with @parente, and we see in `bin/www` that it looks like the https server will handle this for us by passing in the `app` file.

parente:
Will update to keep the code change that moves the logic to config out of `www`, but will not inject the ssl options into the proxy server.

I'm miffed about why I was seeing `POST /api/kernels?1459777914810 - - ms - -` earlier. Could be same rare problem as #156 ?

parente:
@dalogsdon Kept the config cleanup but removed use of the SSL options on the `createProxyServer` call. Added a comment about why it's not config'ed there.

PR #170, opened 2016-04-07
[issue: 168] Option to pass user auth to kernel

### LINKED ISSUE #168: Option to expose auth bundle to kernel gateway
The auth passportjs strategies currently store arbitrary metadata in the express session. In some use cases, the kernel may need information from this bundle in order to act on behalf of the authenticated user. For example, if Dropbox is the configured OAuth dashboard provider, and a dashboard / notebook wants to read/write data from the authenticated dashboard user's Dropbox account, the scoped Dropbox access token granted privileges by the user at login needs to be shared with the kernel.

The kernel gateway has the ability to receive environment variables prefixed with `KERNEL_` on a `POST /api/kernels` request and set these in the environment of a newly launched kernel. An option here in the dashboard server can take advantage of this to:

1. Grab the `user` object from the passport strategy `done(null, user)` call.
2. Pass it along as a JSON value for a `KERNEL_USER_AUTH` env var in the POST request.
3. Write notebooks that read the env var, parse the JSON, and do what they want.

The option on the dashboard server in `config.json` can be:

```
// Forward auth strategy user profile to every kernel launched for every dashboard visited on the server
// e.g., so the kernel can re-use an auth token to access an external service
KG_FORWARD_USER_AUTH: false
```

parente:
Damn near impossible to modify the outgoing body of a proxied request which is what #168 requires. So ...

* Move kernel creation logic to a standard route
* Move all special case tracking performed after kernel creation to that route
* Inject auth info if `KG_FORWARD_USER_AUTH` is set (PS: don't assign to a module global like the others since we want to vary it during testing)
* Upgrade kernel gateway to 0.5.0 for env var support
* Add a test notebook/dashboard that print all env vars for manual testing
* Add unit tests

Fixes #168

### DIFF
--- a/routes/api.js
+++ b/routes/api.js
@@ -7,6 +7,7 @@ var debug = require('debug')('dashboard-proxy:server');
 var error = require('debug')('dashboard-proxy:server:error');
 var Buffer = require('buffer').Buffer;
 var BufferList = require('../node_modules/websocket/vendor/FastBufferList');
+var bodyParser = require('body-parser');
 var WebSocketFrame = require('websocket').frame;
 var nbstore = require('../app/notebook-store');
 var config = require('../app/config');
@@ -15,6 +16,7 @@ var request = require('request');
 var url = require('url');
 var urljoin = require('url-join');
 var urlToDashboard = require('./url-to-dashboard');
+var router = require('express').Router();
 
 var kgUrl = config.get('KERNEL_GATEWAY_URL');
 var kgAuthToken = config.get('KG_AUTH_TOKEN');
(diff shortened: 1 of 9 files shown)


parente:
🍰 for @jtyberg for suffering through this with me.

parente:
@jhpedemonte @dalogsdon can you give this a review and merge?

routes/api.js
```diff
@@ -238,58 +240,99 @@ var removeSession = function(sessionId) {
     return delete sessions[sessionId];
 };
 
-var proxyRoute = function(req, res, next) {
-    proxy.web(req, res);
+// Create a kernel. Does not use the proxy instance because it must parse the
+// and reserialize the request body with additional information in certain
+// configurations.
+router.post('/kernels', bodyParser.json({ type: 'text/plain' }), function(req, res) {
```
jhpedemonte: A little apprehensive about using the Express router within this file. I like keeping that just to *app.js*.

We could do a check in `proxyRoute` to see if the `req` is for `/api/kernels` and if so, do this block of code. See deleted block in `proxy.on('proxyRes', ...`

parente: @jtyberg and I tried numerous combinations of trying to keep everything as middleware, invoke `web.proxy` for all routes except `/api/kernels`, and invoke our special handling for `/api/kernels`. That approach grew really really complex because of the body parser middleware, modifications to the body, and restreaming of the modified body required for the proxy to them forward the modified to the backend. It was quite lengthy and we never did get it working. I think we were fighting a losing battle because all of the examples about intercepting the request body before proxying treat the body as immutable. In our case here, we want to modify the body to inject the auth info.

In the end, we went with having a plain old route handler for this one case where we want to mutate the request body.

jtyberg: This version also has the benefit of having all the magic stuff that needs to happen when creating new kernels all in one spot.

dalogsdon:
I tried out the example notebook and see the KERNEL_USER_AUTH containing my username. :+1:

parente:
I've also tested this with auth-box.js for kicks and have confirmed the two tokens appear in the kernel environment.

jhpedemonte:
Code looks good to me.

PR #173, opened 2016-04-11
[issue: 148] Update readme about deployment

### LINKED ISSUE #148: Need better instructions for bundling associated resources
We need better instructions for how to use the associated resources feature of contentmanagement/bundler extensions. Currently, the bundler has a single sentence pointing to a demo notebook in the contentmanagement extension.

* [x] Add clear, concise instructions for how to use the associated resources feature. Not sure where it belongs (cms, bundler?), but it should definitely be referenced in cms, bundler and dashboard server repos.
* [x] Update cms' associations demo. When viewed on GitHub, the notebook is rendered, therefore a user cannot see the HTML comment convention for specifying resources. The only option is to view it *raw*, which makes it difficult to copy/paste. Better to also duplicate the syntax in a markdown comment, so it is properly rendered in GitHub or Dashboards Server.

parente:
* Extend diagram to include notebook server
* Add workflow steps that we're trying to enable
* Add reference to all-in-one gist
* Add reference to associations tutorial

Fixes #148


parente:
@jhpedemonte for your review

PR #179, opened 2016-04-13
[issue: 178] Immediately load index/index.ipynb bundles

### LINKED ISSUE #178: index bundles don't act like index.ipynb files
1. Put an index.ipynb in the root of the dashboard server.
2. Any visit to / shows the dashboard.
3. Replace it with an index folder / bundle with an index.ipynb 
4. The dashboard listing shows.

Expected behavior: the bundled dashboard behaves like index.ipynb

Why:

The main use case for a chromeless server is to show a particular dashboard without making the user aware that it's from a jupyter dashboard server. When hitting the root of the server (or any other path that isn't a known dashboard), the user gets a:

```
Error 404
List not supported
```

which is ugly. It'd be much more natural to have an index folder at `/` display immediately like on any other website.

parente:
* If a directory contains a folder named index which itself contains an index.ipynb, redirect to it immediately in place of showing a directory listing (i.e., makes bundled dashboards named index behave like unbundled dashboards named index.ipynb)
* Allow access to /dashboards/* under /* by adjusting route ordering and patterns slightly (previously, only the root (/) listing or index.ipynb was accessible without /dashboards)
* Comment the hell out of the renderer.render logic

Fixes #178

### DIFF
--- a/routes/renderers.js
+++ b/routes/renderers.js
@@ -13,8 +13,9 @@ var path = require('path');
 var Promise = require('es6-promise').Promise;
 var urljoin = require('url-join');
 
+var DB_INDEX = config.get('DB_INDEX')
+var DB_INDEX_DIR = config.get('DB_INDEX_DIR')
 var DB_EXT = config.get('DB_FILE_EXT');
-var INDEX_NB_NAME = config.get('DB_INDEX');
 var reNbExt = new RegExp('\\' + DB_EXT + '$');
 
 function _renderList(req, res, next) {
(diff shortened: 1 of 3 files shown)


routes/renderers.js
```diff
@@ -108,40 +109,62 @@ function _render(req, res, next, opts) {
...
+                var dbpathIndex = path.join(dbpath, DB_INDEX_DIR, DB_INDEX);
+
+                // If the path exists on disk and is a normal directory, check
+                // if it contains an index bundle.
+                nbstore.stat(dbpathIndex)
+                    .then(function() {
+                        // If it does contain an index bundle, redirect to it.
+                        // We can't render it from because requests for static
+                        // assets from the frontend will fail (off by one path).
+                        res.redirect(path.join(req.path, DB_INDEX_DIR));
```
dalogsdon: Since this is a URL, do you want to use `urljoin` instead of `path.join`?

routes/renderers.js
```diff
@@ -108,40 +109,62 @@ function _render(req, res, next, opts) {
...
+                            var err = new Error('List not supported');
+                            err.status = 404;
+                            next(err);
+                        } else {
+                            _renderList(req, res, next);
+                        }
+                    });
             } else {
-                // plain file -- return contents
+                // If the path is neither a dashboard or a directory, send it
```
dalogsdon: neither...nor? (sorry)

parente: Fixed.

routes/renderers.js
parente: Fixed.

dalogsdon:
Looks good, works for me. :+1: for the renderer comments

parente:
🎈

PR #180, opened 2016-04-13
[WIP] [Issue 97] Use higher-level WebSocketConnection instead of *Frame

jhpedemonte:
Should handle large messages that are sent across multiple WS frames (which we don't currently handle and instead just error out).

This does allow WS messages: client <> proxy <> kernel gateway. However, not currently doing message rewriting.

Ref #97

Issue #97, opened 2016-02-16
Handle large websocket messages spread out across 2+ messages


jhpedemonte:
If a websocket message is large enough, the data will be sent across 2 or more separate messages. We don't currently handle that in the rewriting code (either with our custom WS code or the library).

We're only rewriting messages in the client -> kernel gateway direction. `execute_request` messages coming from the client should be small, since the `code` section will just be an index. Therefore, there are 2 scenarios (at least) that we need to worry about:

1. When replacing the the index with actual code, that increase the size of the message. Of course, the max size for a message is around 1MB (I think), so the chances of there being a code block larger than that are pretty small. But it is still a possibility.
2. Ensure that any other messages are properly passed through without modification. That way, if there is data that is too large and is spread across multiple messages, the server should just pass it through.

parente:
Possible that #156 is related to this? Maybe the widget data messages are large when there are a lot of registrations in the streaming demo?

jhpedemonte:
Server will output the error `insufficient data for frame` if this is the case (when logging is turned on). Have you seen this in #156?

parente:
Haven't tried to reproduce with logging on.

parente:
With logging on, yes, I see the insufficient message. The logs are in #156.

parente:
> We're only rewriting messages in the client -> kernel gateway direction. execute_request messages coming from the client should be small, since the code section will just be an index. 

This is true, but it assumes we're not touch messages from kernel gateway to client. But the log messages show we definitely are still building WebSocketFrame object for messages flowing from KG -> client, even though we don't even modify them because they're not execute_requests. This is proven by the presence of `PROXY: received message from client WS` in the log with comm channel payloads that are responses from the kernel.

We need to be smarter about which messages we touch in `socket.emit` which appears to get invoked with data flowing in **either** direction.

parente:
Another, probably related problem, is that the bufferList object and others are globals in the api.js module. What happens when two websockets are connected and data needs to be retained in the bufferList across data events?

parente:
Few facts to note:

1. The `socket.emit` for `eventName=data` fires for both client to kernel and kernel to client messages.
[...]

PR #181, opened 2016-04-13
[Issue 97] Use higher-level WebSocketClient/Server classes, instead of *Frame

jhpedemonte:
Use `WebSocketServer` and `WebSocketClient` to do WS proxying, instead of relying on `http-proxy`.

Ref #97
Alternate to #180


jhpedemonte:
This is a different take than #180. Instead of using `http-proxy` to do websocket proxying, this attempts to create a WS server (which connects to browser) and a WS client (which connects to kernel gateway) and proxy data between the two.  If this worked, it would be relatively easy to add in our message rewriting.

However, it's not working for me. I don't see any messages coming in (when running `make dev-logging`).

jhpedemonte:
Sweet! Latest update makes things work. Tested on all of our example notebooks, although didn't run tests yet.  Still needs lots of cleanup.

@parente, please test to see if it fixes issue you were seeing with messages split across multiple WS frames.

parente:
Will do.

parente:
tl;dr: Works great.

Deployed the streaming meetup notebook to a dashboard server. Opened a browser tab to it. Started the stream. No problems for 5 minutes.

Decided to push my luck. Opened a second browser tab to the same dashboard. Started another stream. Walked away for 1.5 hours. My laptop may have suspended itself at some point. It at least locked itself.

Came back. Checked the consoles. No errors. One of them still streaming away happily. The other one apparently stopped. Toggle the stream on / off on that one got it going again immediately.

Websocket problem is gone.

parente:
I know this PR isn't done, but one reminder: the kernel cleanup logic needs to be added back in. Probably got axed when the http proxy bit went away. It's probably significantly easier to implement now given that the websocket server probably emits a close/disconnect event when the browser goes away.

routes/api.js
```diff
@@ -84,24 +88,103 @@ var substituteCodeCell = function(payload) {
...
+        });
+
+        client.on('connectFailed', function(e) {
+            error('ws client failure', e);
+        });
+        client.on('error', function(e) {
+            error('ws client error', e);
+        });
+        client.on('close', function() {
+            debug('closing WS client connection');
```
parente: Probably need to move the kernel deletion timer and callback here. Doesn't seem to fire anymore on server close.

routes/api.js
```diff
@@ -84,24 +88,103 @@ var substituteCodeCell = function(payload) {
...
+        // XXX verify request.origin
+        var servConn = request.accept(null, request.origin);
+
+        var pendingServMsgs = [];
+        var bufferServMsgs = function(data) {
+            pendingServMsgs.push(data);
+        };
+        servConn.on('message', bufferServMsgs);
+        servConn.on('close', function(reasonCode, desc) {
+            debug('closing WS server connection:', reasonCode, desc);
```
parente: Probably need to move the kernel deletion timer and callback here. Doesn't seem to fire anymore on server close.

routes/api.js
```diff
@@ -84,24 +88,103 @@ var substituteCodeCell = function(payload) {
...
+
+        var wsKgUrl = urljoin(kgUrl, kgBaseUrl, request.resourceURL.path).replace(/^http/, 'ws');
+        client.connect(wsKgUrl, null);
+    });
+
+    return;
+
+
     // Listen to the `upgrade` event and proxy the WebSocket requests as well.
     _server.on('upgrade', function(req, socket, head) {
```
parente: I'm assuming this part goes away now? Are we sure it wasn't active somehow during the testing?

jhpedemonte:
> I know this PR isn't done, but one reminder: the kernel cleanup logic needs to be added back in

Yep, Working on refactoring and making this pretty!

routes/api.js
jhpedemonte: Yes, goes away. The `upgrade` is now handled internally by the `WebSocketServer/Client` classes.

jhpedemonte:
Latest commit passes integration tests. But unit tests have to be rewritten.

app/ws-rewriter.js
```diff
@@ -0,0 +1,219 @@
...
+
+    wsserver.on('request', this._handleWsRequest.bind(this));
+}
+util.inherits(WsRewriter, EventEmitter);
+
+WsRewriter.prototype._handleWsRequest = function(req) {
+    debug('ws-server connection request');
+
+    var self = this;
+    var servConn = req.accept(null, req.origin); // accept from all origins
```
parente: Might need to think about XSRF if we're passing any origin through.

routes/api.js
```diff
@@ -40,164 +36,50 @@ var proxy = httpProxy.createProxyServer({
...
-                        return ''; // throw away execute request
-                    });
-            }
-        } catch(e) {
-            // TODO better handle parse error in WS message
-            error('Failed to parse `data` in WS.', e);
-            transformedData = '';
-        }
+function initWsProxy(server) {
+    if (wsProxy) {
```
parente: Are we still sharing one instance of the rewriter here?

parente:
Tested with the docker-compose recipe by running 2 instances of the streaming meetup demo simultaneously. No websocket disconnects or other problems noted. Ran all the unit and integration tests and no problems there either.

Few minor questions inline above. Otherwise, good stuff. 🏆  Let's :shipit:

app/ws-rewriter.js
jhpedemonte: I was wondering about that, but not sure what we would do here. This is the websocket server, so it has to accept incoming connections. It can't really know ahead of time who the client will be -- unless you have something else in mind?

routes/api.js
jhpedemonte: I think we should be fine. There is only one instance, but it sets up the websocket server such that on each `request`, it creates a new client to the backend and then proxies those connections. And each new connecting client generates a new `request`.

I think I tested this early at one point, but would need to do so again with latest code.

app/ws-rewriter.js
parente: Nothing to solve here. We'd solve it at the auth level. Something akin to https://github.com/expressjs/csurf where we render a token onto the HTML page of the dashboard and expect to see that token back in the request.

We might already be getting something like XSRF protection from the way we're using session ID. Another security review is probably in order. For now, leave it as-is.

routes/api.js
parente: I tested with multiple clients and studied the code a bit more after I commented. I believe you're right that the only singleton is the WS server that accepts connections and spawns of instances of client/server to handle them independently. I couldn't see in the code any state that would be incorrectly shared across requests nor could I find a case in testing that causes crosstalk across clients. 👍

Issue #141, opened 2016-03-14
Some dashboards not working in Safari


dalogsdon:
Simple dashboards are working fine in Safari, but the plotting demo and resizable widget demo do not.

There are no obvious relevant errors in Javascript console.

Steps to complete:

* [ ] Move to ipywidgets 5.1.x + declarative widgets 0.5.2
* [ ] Either go with #108 or move to latest jupyter-js-notebook
* [ ] Either make rendermime's HTML transform work in Safari or replace it here with one that does

nitind:
WIth a **lot** of help from @dalogsdon, determined that a DOMException was being silently ignored in the transformime HTML code because a Promise lacked a reject callback.


https://github.com/nteract/transformime/blob/master/src/html.transform.js#L5 encounters NotSupportedError: DOM Exception 9 in Safari.
https://github.com/nteract/transformime/blob/master/src/transformime.js#L70 catches it and calls Promise.reject(), but https://github.com/jupyter/jupyter-js-notebook/blob/7b56af00451435caf3e39681571871471fce7c7e/src/output-area/widget.ts#L254 has no "reject" callback to handle or log the exception.

A lot of this was refactored right around the time we started working on this, so it's possible that the problem has been fixed by the switch to rendermime (specifically https://github.com/jupyter/jupyter-js-ui/blob/master/src/renderers/index.ts#L31 no longer calling `Range#createContextualFragment()`), but moving the dashboard server to the newer version of jupyter-js-notebook appears non-trivial.

parente:
We need to update to the newest jupyter-js-notebook if it has moved on and is using rendermime instead. No sense in opening a defect against a lib that is no longer used.

Give the upgrade a shot and let's see how much else needs to change. Tap @jasongrout for help on the new output area usage as needed.

jhpedemonte:
Worked with @nitind to update our code to use latest `jupyter-js-notebook`. One big issue that hit us was needing to set `outputAreaModel.trusted = true`, otherwise plotting libs JS (such as from **matplotlib**) won't run in the browser. This got us working with ipywidgets and matplotlib.

But **Bokeh** and **Plot.ly** are still not rendering.  I see 2 different JS blocks added to the DOM for Bokeh. I inspected the websocket frames:  the first incoming JS is large (191224) and includes **bokeh.min.js**. It ​does​ get executed -- I see the console logs.

[...]

PR #182, opened 2016-04-13
[Issue: 141] Upgrade jupyter-js-* dependencies

nitind:
Update versions of jupyter-js-notebook and jupyter-js-services. Fix
    for Safari in still newer jupyter-js-notebook introduces a blocker in
    Chrome.
(c) Copyright IBM Corp. 2016


nitind:
For https://github.com/jupyter-incubator/dashboards_server/issues/141 .

parente:
`test/test_bundled` appears to have stopped working with this PR.

parente:
... possibly because we now need to be on declarativewidgets==0.5.0dev1

parente:
Same with `demo/taxi_demo`, so, yes, appears to be isolated to demos using declarative widgets.
@jhpedemonte how did you go about bundling these originally so @nitind can do the same?

jhpedemonte:
I think I used the all-in-one gist, did a deploy, then pulled out the files from the Docker container. Not the best solution. But easiest would be to use the bundler, now that there's a "download as zip" option for the dashboard server.

parente:
@nitind if you make dev-with-dashboard-server over in dashboard_bundlers and drop those two notebooks in, you can get the zip @jhpedemonte is talking about.

Of course, I'm just guessing this is why declarative widget demos are failing. Could be there's new incompatibilities as well.

parente:
Tested with @nitind:

1. With this PR in place, and decl widgets 0.4.5, we see an error on page load stating:

```
Uncaught Error: Script error for "nbextensions/widgets/widgets/js/manager", needed by: nbextensions/urth_widgets/js/widgets/DeclWidgetModel
http://requirejs.org/docs/errors.html#scripterror
```

2. With this PR in place, and decl widgets 0.5.0.dev1, we get no errors, but initialization does not occur either and so no widgets appear on the page.

3. On **master** with decl widgest 0.5.0.dev1, everything works as expected.

So the incompatibility has something to do with decl widgets and the jupyter-js-notebook (and its 2nd order dependencies) version bump in this PR.

nitind:
With 0.5.0.dev1, the `Kernel.start` promise gives us a `kernel` whose `status` is 0, "Unknown". Our shimmed `is_connected()` function returns `false` at that moment .The declarative widget `created` function relies on either `kernel_ready.Kernel` firing after it's added its listener, or `is_connected()` returning `true`, and our setup we've already fired the event and our dashboard.js `is_connected()` is still returning `false`.

nitind:
Amended to still support declarative widgets.

public/js/dashboard.js
```diff
@@ -111,6 +108,8 @@ requirejs([
                 widgetManager.trackPending(kernelFuture, $widgetSubArea.get(0), model);
             });
         });
+    }, function(err) {
+        console.error('Failed to start kernel:', err);
```
jhpedemonte: Is there anything else we want to do here, rather than just log?

Since this is such a serious error, we should probably show an error page/dialog to the user explaining the situation (could be handled as separate issue).

public/js/dashboard.js
```diff
@@ -139,8 +138,20 @@ requirejs([
...
+                    nb.kernel.statusChanged.disconnect(puller);
+                    nb.events.trigger('kernel_ready.Kernel');
+                }
+            }
+            nb.kernel.statusChanged.connect(puller);
+        }
+        else {
+            // kernel has already started
+            nb.events.trigger('kernel_ready.Kernel');
+        }
```
jhpedemonte: Good catch!

public/js/dashboard.js
parente: I think open that as a separate issue. Clearly we can be doing better error handling across the board.

public/js/dashboard.js
nitind: Well, I had help.

parente:
Thanks @lbustelo for helping @nitind here. 

I'm marking this as HOLD because it makes the dashboard server dependent on a pre-release of declarative widgets. When decl widgets 0.5.0 comes out, we can merge.

@nitind I'm assuming the timing fix for the widgets does not also by chance fix the Safari issue, right?

jhpedemonte:
The `kernel_ready` change can definitely go in now.

parente:
>The kernel_ready change can definitely go in now.

Yea, this is true. We probably have this race condition today w/ 0.4.5 as well. Wonder if you can separate out that bit into another PR, @nitind.

parente:
> I'm marking this as HOLD because it makes the dashboard server dependent on a pre-release of declarative widgets.

Turns out we're already dependent on 0.5.0. The decl widgets are backward compatible with the old init function signature, but the dashboard.js invocation of that function and the dependency on the newer ipywidgets mean the code in master is incompatible with decl widgets < 0.5.0.dev0.

parente:
Subsumed by #187.

Issue #157, opened 2016-03-24
Support report layout


dalogsdon:
https://github.com/jupyter-incubator/dashboards/issues/203 will introduce a simpler dashboard "report" layout:

> **Report mode:** a simple view which shows cells in notebook order, single column and full width. The only action available to the user is whether to show/hide each cell.

We should support this layout once it is implemented in the dashboards notebook extension.

dalogsdon:
Discussed with @jhpedemonte, and we decided that https://github.com/jupyter-incubator/dashboards/issues/203 will be handled as a subset of the grid layout. Therefore, this should just automatically work in the dashboard server. We will probably just end up closing this issue after verifying that it works as expected.

parente:
Is the dashboard authoring for report mode computing height for each cell? Or will logic need to be added over here to interpret and compute the height of each output area?

I ask because #145 is about rendering notebooks that have NO layout metadata as if they were laid out top to bottom in report mode with all cells with output showing.

dalogsdon:
Yes, it should still be computing height.

dalogsdon:
The nature of jupyter-incubator/dashboards#203 has changed and is now handing report layout as a separate (non-Gridstack) layout. The dashboard server will have to handle this. This may now overlap with #145 a bit as the dashboard will default to the report layout if no dashboard layout metadata.

parente:
Probably best if @dalogsdon and @jameslmartin pair on the two items in tandem. Good intro for @jameslmartin too.

PR #183, opened 2016-04-13
[Issue 145] Fix notebook dashboards with no layout in metadata

### LINKED ISSUE #145: "no layout" dashboards handled incorrectly
I think we are handling notebooks without layout metadata incorrectly. Take a look at our **test_no_layout.ipynb** notebook. When opened in the dashboard server, we show nothing.

The dashboard extension runs through an algorithm for notebooks without layout metadata, determining which cells contain output (and should therefore be shown) and hiding others. But such an algorithm doesn't make sense for dashboard server.

Instead, I propose that we show all of the cells in a single-column, full width (if no notebook-level `metadata.urth.dashboard` property is set). We should also log a message to the browser console. This works as a quick way to upload and display a notebook, without worrying about installing/using the dashboard & bundler extensions.

jameslmartin:
Fixes #145

### DIFF
--- a/less/dashboard.less
+++ b/less/dashboard.less
@@ -51,8 +51,8 @@ body {
             img {
                 max-width: 100%;
                 height: auto;
-                // margin-left: auto; // possible solution to #146
-                // margin-right: auto;
+                margin-left: auto;
+                margin-right: auto;
             }
 
             .widget-subarea:empty {
--- a/less/dashboard.less
+++ b/less/dashboard.less
@@ -51,10 +51,18 @@ body {
             img {
                 max-width: 100%;
                 height: auto;
+                // margin-left: auto; // possible solution to #146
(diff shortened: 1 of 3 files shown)


parente:
@dalogsdon should review this one for sure. 🏄

dalogsdon:
This looks fine. I noticed a couple minor styling issues (non-centered layout and padding on empty cells). However I'm ready to begin #157 which will be touching the same code so I will go ahead and merge this.

dalogsdon:
I noticed an issue I missed when reviewing this: markdown cells won't render in no-layout notebooks. The cell needs the content `{{{markdownContent cell_type source}}}` in order to render the markdown client-side. We don't need to fix this here as I've already fixed it during my work on #157 so it will be resolved soon; this is just an FYI.

PR #187, opened 2016-04-16
[WIP] [issue: 141] Safari fix (but at a cost)

parente:
Builds on #182 as an experiment.

Fixes:

* All demos (including ones with declarative widgets) except test_plotting_libs work in Safari, Firefox, and Chrome

How:

* Build a transformime object with the standard set of transformers

New issues:

* Niether bokeh nor plot.ly work (unknown reason)
* Execute results show output area markings that either need to be removed or styled away (see the big cats demo, some Out[x] nodes in there )
* Output DOM node classes are more Jupyter Lab-centric than classic notebook centric (see the env vars demo, the printed text is not styled as monospace)

As an alternative to manipulating the output via styling, we could implement our own OutputAreaWidget to control how results are placed in the DOM independent of how the JLab notebook evolves.


parente:
Updated this to get bokeh working with a hack. Short of it is that rendermime does not eval scripts in HTML which is what bokeh is using. 

Leaving this as WIP since I think we want to go another route. Will put the details in #141.

parente:
@jhpedemonte I'm going to close this because I expect that you'll be submitting another PR to address #141 either by using your own output area for #191 or the latest jupyter-js-notebook.

PR #190, opened 2016-04-18
[Issue 157] Support report layout

### LINKED ISSUE #157: Support report layout

dalogsdon:
Resolves #157

### DIFF
--- a/etc/notebooks/test/test_style_match.ipynb
+++ b/etc/notebooks/test/test_style_match.ipynb
@@ -257,7 +257,7 @@
     }
    },
    "source": [
-    "<img src=\"http://www.bigcat.com/CZ2/cougar2.jpg\">"
+    "<img src=\"http://cincinnatizoo.org/wp-content/uploads/2011/02/cougar1.jpg\">"
    ]
   },
   {
@@ -328,6 +328,7 @@
   },
   "urth": {
    "dashboard": {
+    "layout": "grid",
     "cellMargin": 10,
     "defaultCellHeight": 20,
     "layoutStrategy": "packed",
--- a/routes/renderers.js
+++ b/routes/renderers.js
@@ -13,9 +13,10 @@ var path = require('path');
 var Promise = require('es6-promise').Promise;
 var urljoin = require('url-join');
 
-var DB_INDEX = config.get('DB_INDEX')
-var DB_INDEX_DIR = config.get('DB_INDEX_DIR')
+var DB_INDEX = config.get('DB_INDEX');
+var DB_INDEX_DIR = config.get('DB_INDEX_DIR');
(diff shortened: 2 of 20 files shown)


parente:
Hold until jupyter-incubator/dashboards#211 goes in.

etc/notebooks/test/test_style_match.ipynb
```diff
@@ -257,7 +257,7 @@
     }
    },
    "source": [
-    "<img src=\"http://www.bigcat.com/CZ2/cougar2.jpg\">"
+    "<img src=\"http://cincinnatizoo.org/wp-content/uploads/2011/02/cougar1.jpg\">"
```
parente: This was bothering me. Thank you.

routes/renderers.js
```diff
@@ -85,12 +85,12 @@ function _renderDashboard(req, res, next, opts) {
...
-                return ( currentCell.hasOwnProperty('metadata') &&
-                         currentCell.metadata.hasOwnProperty('urth') &&
-                         currentCell.metadata.urth.hasOwnProperty('dashboard') );
-            });
+            // get dashboard layout from notebook-level dashboard metadata
+            var dashboardLayout = (notebook.metadata &&
+                                  notebook.metadata.urth &&
+                                  notebook.metadata.urth.dashboard &&
+                                  notebook.metadata.urth.dashboard.layout) ||
+                                  'no-layout';
```
parente: We decided no layout metadata meant render as a report here. But in the extension we said if layout was missing, we'd look for grid metadata and assume grid is the default. Hmmm.

If I take a notebook that was produced by the 0.4.x extension and drop it in dashboard server now with this change, it's going to render in report layout and ignore all the grid metadata. That seems ... unfriendly.

Can we at least do a sniff test to see if any of the cells have the `urth.dashboard` path. If any of them do but layout is undefined in the notebook level metadata, default to grid. If there's no urth.dashboard anywhere then take the no-layout route.

dalogsdon: I agree, and this is what @jameslmartin did in #145. I can reinstate that check.

parente: OK. Cool. Just ping again when it's here and I'll finish testing.

dalogsdon:
I updated the renderer to use grid layout if no layout was set but there is at least 1 cell that contains the grid metadata (row, col, width, height). I also noticed the Config script code in the dashboard template was always setting grid layout properties and not always putting in all default values if only some were set in the notebook. I've updated the template to handle these cases.

Issue #171, opened 2016-04-07
Use cookie session storage


parente:
Currently using MemoryStore for sessions. Fine for dev. At deploy time, want something more persistent. Since we store very little, http://www.senchalabs.org/connect/cookieSession.html should do.

PR #197, opened 2016-04-20
[issue: 171] Use cookie session storage

jameslmartin:
Implements #171 to store session data using Express cookie-sessions

(c) Copyright IBM Corp. 2016

### DIFF
--- a/app.js
+++ b/app.js
@@ -7,6 +7,7 @@ var session = require('express-session');
 var path = require('path');
 var logger = require('morgan');
 var cookieParser = require('cookie-parser');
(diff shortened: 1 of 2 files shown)


dalogsdon:
Do we still need to be using `express-session` now that we are using `cookie-session`? I don't think I see any references to it anymore in app.js.

parente:
If it's still in the package.json, yea we can remove it.

dalogsdon:
It is in package.json and required in app.js with no references to it.

parente:
@jameslmartin can you submit a follow-on PR?

PR #198, opened 2016-04-21
Fix KG authorization header missing

parente:
Missed in review of PR #181


parente:
@jhpedemonte can you give this a quick review and merge?

jhpedemonte:
I think this should be more generic. `ws-rewriter.js` should just take a `headers` argument which it sends along. Then the auth header is created in `api.js` and passed along when doing `new WsRewriter({...})`.

parente:
Updated to form the headers in api.js and pass them in.

PR #200, opened 2016-04-26
[Issue 194] Upgrade to ipywidgets 5.0

### LINKED ISSUE #194: Move to ipywidgets 5.0.0 / jupyter-js-widgets 1.0
Maybe move to 5.0.0b4 now so that the jump to final is easy. Or wait for the final releases of both.
At least one recent change that we might need to deal with:

* https://github.com/ipython/ipywidgets/pull/461 (ref'ed by https://github.com/ipython/ipywidgets/issues/522)

jhpedemonte:
Updates to ipywidgets 5.0 and fixes loading issues in the browser by compiling all of our code into one large file (> 2MB!) using webpack. Minification can reduce this to ~800K, but adds time to the build -- so not something we want to do during development.

Unit & integration tests and all demos are working.

**Todo:**
* [x] Dynamic rebuild during development is broken. Need to rerun `make dev` after code change.
* [x] Distinguish between DEVELOPMENT and PRODUCTION build. That way, we can enable **minification** and **source-map** during the latter.
* [x] Move our code out of `/public/js`. Change built artifact to go into that directory (and get rid of `/public/components`).

Fixes #194


jhpedemonte:
@parente I think this one is ready to go.

client/js/layout.js
```diff
@@ -28,7 +28,6 @@ define([
         }).get().reduce(function(a, b) {
             return Math.max(a, b);
         }, 0);
-    var height = maxY * cellHeight + (maxY - 1) * cellMargin + maxHeight;
```
parente: Was this a leftover?

parente:
Tested lots of notebooks manually in `make dev` and `make demo-container` (minified). Ran unit, integration, and install tests.

We'll need to update the README about what libs we're compatible with before tagging 0.5.0. Other than that, :shipit: !

Issue #169, opened 2016-04-06
Frontend hardcodes python3


parente:
Need to make the decision about what language to use server-side after loading the notebook. That, or the server has encode the kernel name in the page template and let the JS echo it back. 

I don't know which one is better / worse yet. 💭

parente:
@jameslmartin want to take a crack at this one next?

Since this was opened, code was added to modify `POST /api/kernel` requests before they reach the kernel gateway. The fix here would be to get the kernel name (language) from the notebook metadata and replace the `{"name" : "kernel_name"}` value in the POST request from the frontend with it before it gets proxied to the KG.

Issue #203, opened 2016-04-28
Example R / Scala notebooks


parente:
Need some notebooks in `test/` for manually testing that R and Scala both work properly.

* ~~[ ] R notebook that just prints~~ See comment about R support
* ~~[ ] Scala notebook that just prints~~(Should go in #211 for simple manual testing)
* ~~[ ] R notebook that uses minimal declarative widgets~~ See comment about R support
* [x] Scala notebook that uses minimal declarative widgets (This is now the one and only goal)

Don't want to add massive new demos that we need to maintain. Just want a sanity check of how these things behave in dashboard server.

Needs to come after #169

parente:
https://github.com/IRkernel/IRkernel/pull/312 fixes the R kernel so that it works with jupyter-js-services. Not sure when there will be a release with it, so for the time being, focus on Scala and Toree.

parente:
Description updated. Moved this out of 0.6.0.

parente:
We did not check / install the necessary Toree declarative widget deps in #169. They might already be installed properly in the Dockerfile.kernel or they might not. If not, that work will need to be done as part of this issue.

jameslmartin:
We have a choice to make concerning the declarative widgets .jar file for Scala. Currently, we ask the author to specify the IP and port of the notebook server to successfully load the .jar file. 
```scala
// modify to IP and Port of this notebook server
%addjar http://localhost:8888/nbextensions/declarativewidgets/declarativewidgets.jar

import declarativewidgets._

initWidgets
```
This introduces a problem when deploying the notebook to a dashboard server, as the dashboard server should stand alone and does not rely on a notebook server.

@parente made one suggestion that we could move the .jar file to be public facing, but we may then run into problems with versioning. We could also load the .jar file using the path to where the file is in the local file system of the Docker notebook server container. This doesn't quite solve the portability issue unless we put the .jar file along the same path in the dashboard server.

Any other suggestions or workarounds?

parente:
@lbustelo, @jhpedemonte any thoughts on how the declarativewidgets.jar should be referenced kernel-side so that it works both in notebook server and in kernel gateway (deployed)?

An analogous `import foo` in Python works because we assume the notebook environment and the KG environment both have package `foo`. The `addjar` is more complicated because an absolute file or URL has to be given. (Unless it searches the classpath too?)

jameslmartin:
[...]

PR #211, opened 2016-05-04
[Issue: 169] Frontend hardcodes python3

jameslmartin:
This resolves the language of the kernel before sending the POST request to the kernel gateway (#169) . Though the request body is successfully set to the correct language, no activity is observed over the WebSocket channels.

### DIFF
--- a/Dockerfile.kernel
+++ b/Dockerfile.kernel
@@ -1,7 +1,7 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-FROM jupyter/pyspark-notebook:0017b56d93c9
+FROM jupyter/all-spark-notebook:0017b56d93c9
 
 RUN pip install jupyter_kernel_gateway==0.5.0
 # install ipywidgets 5.x
--- a/client/js/widget-manager.js
+++ b/client/js/widget-manager.js
@@ -10,7 +10,7 @@
 var $ = require('jquery');
 var Widgets = require('jupyter-js-widgets');
 
-    var WidgetManager = function(kernel, msgHandler) {
+    var WidgetManager = function(kernel, kernelname, msgHandler) {
         //  Call the base class.
         Widgets.ManagerBase.call(this);
 
@@ -24,17 +24,20 @@ var Widgets = require('jupyter-js-widgets');
         // Register the comm target
         this.commManager.register_target(this.comm_target_name, this.handle_comm_open.bind(this));
 
-        // Validate the version requested by the backend.
-        var validate = (function validate() {
-            this.validateVersion().then(function(valid) {
-                if (!valid) {
-                    console.warn('Widget frontend version does not match the backend.');
-                }
-            }).catch(function(err) {
-                console.error('Could not cross validate the widget frontend and backend versions.', err);
-            });
-        }).bind(this);
-        validate();
+        // Validate the version requested by the backend -- required for ipywidgets
+        kernelname = kernelname.toLowerCase();
+        if (kernelname === 'python3' || kernelname === 'python2') {
+            var validate = (function validate() {
+                this.validateVersion().then(function(valid) {
+                    if (!valid) {
+                        console.warn('Widget frontend version does not match the backend.');
+                    }
+                }).catch(function(err) {
+                    console.error('Could not cross validate the widget frontend and backend versions.', err);
+                });
+            }).bind(this);
+            validate();
+        }
 
         this._shimWidgetsLibs();
     };
--- a/client/js/widget-manager.js
+++ b/client/js/widget-manager.js
@@ -25,8 +25,9 @@ var Widgets = require('jupyter-js-widgets');
         this.commManager.register_target(this.comm_target_name, this.handle_comm_open.bind(this));
 
         // Validate the version requested by the backend -- required for ipywidgets
+        // If kernel name is not defined, default to python3
         kernelname = kernelname.toLowerCase();
-        if (kernelname === 'python3' || kernelname === 'python2') {
+        if (!kernelname || kernelname === 'python3' || kernelname === 'python2') {
             var validate = (function validate() {
                 this.validateVersion().then(function(valid) {
                     if (!valid) {
--- a/routes/api.js
+++ b/routes/api.js
@@ -14,6 +14,8 @@ var urljoin = require('url-join');
 var urlToDashboard = require('../app/url-to-dashboard');
 var WsRewriter = require('../app/ws-rewriter');
 
+var nbstore = require('../app/notebook-store');
+
 var kgUrl = config.get('KERNEL_GATEWAY_URL');
 var kgAuthToken = config.get('KG_AUTH_TOKEN');
 var kgBaseUrl = config.get('KG_BASE_URL');
@@ -40,19 +42,11 @@ function initWsProxy(server) {
     if (wsProxy) {
         return;
     }
-    
-    var headers = null;
-    if(kgAuthToken) {
-        // include the kg auth token if we have one
-        headers = {};
-        headers.Authorization = 'token ' + kgAuthToken;
-    }
 
     wsProxy = new WsRewriter({
         server: server,
         host: kgUrl,
         basePath: kgBaseUrl,
-        headers: headers,
         sessionToNbPath: function(session) {
             return urlToDashboard(sessions[session]);
         }
@@ -144,6 +138,22 @@ router.post('/kernels', bodyParser.json({ type: 'text/plain' }), function(req, r
              KERNEL_USER_AUTH: JSON.stringify(req.user)
         };
     }
+
+    var notebookPathHeader = req.headers['x-jupyter-notebook-path'];
+    var matches = notebookPathHeader.match(/^\/(?:dashboards(-plain)?)?(.*)$/);
+    if (!matches) {
+        error('Invalid notebook path header');
+    }else{
+        var notebookPath = matches[2];
+        var kernel_lang;
+        var nb = nbstore.get(notebookPath)
+          .then(function success(notebook){
+            console.log("here");
+            kernel_lang = notebook.metadata.kernelspec.name;
+            console.log(kernel_lang);
+          });
+    }
+    console.log(kernel_lang);
 
     // Pass the (modified) request to the kernel gateway.
     request({
--- a/routes/api.js
+++ b/routes/api.js
@@ -139,52 +139,51 @@ router.post('/kernels', bodyParser.json({ type: 'text/plain' }), function(req, r
         };
     }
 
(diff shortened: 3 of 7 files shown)


routes/api.js
```diff
@@ -145,36 +139,56 @@ router.post('/kernels', bodyParser.json({ type: 'text/plain' }), function(req, r
...
+            if(err) {
+                error('Error proxying kernel creation request:' + err.toString());
+                return res.status(500).end();
+            }
+            // Pass the kernel gateway response back to the client.
+            res.set(response.headers);
+            res.status(response.statusCode).json(body);
+          });
+        });
+      }
```
jhpedemonte: I know this is WIP, so I won't have full review comments, but I just want to say that you should adhere to the coding style already in the file:
1. Indentation is 4 spaces (not 2).
2. Add appropriate spaces around keywords (i.e. `} else {`).

parente:
@jameslmartin Focus on trying Toree next since the IRKernel fix might not be released for some time. I know for a fact that Toree works through jupyter-js-services and kernel gateway since it's one of our demos for the kernel gateway project. If you don't see it working here with a simple print, then something else is wrong.

Other than code style review like @jhpedemonte gave above, we're holding off on merging this until we see it work with at least one other kernel beyond Python.

parente:
I see you fixed Toree in your last commit. Cool.

I think this just needs a rebase on master, a fix of the style @jhpedemonte pointed out, and the inclusion of one really simple Scala notebook for manual test and we can send it in.

parente:
@jameslmartin and @jhpedemonte will pair today to see where this stands and how to get it in for 0.6.0

jhpedemonte:
Paired with @jameslmartin to finish this up. Found a couple of things:
1. Kept getting validation error for ipywidgets code. We decided to expose the kernel name in the HTML template and only run the validation for python kernels.
2. Ran across the `Missing property 'buffers'` error. This happens when the Scala kernel craps out due to insufficient memory. Looks like the kernel gateway tries to restart it and sends back websocket messages with `execution_state: "restarting"`. However, those don't have a `buffer` property, which the *jupyter-js-services* code expects (even if empty). Kernel gateway bug?

client/js/widget-manager.js
```diff
@@ -10,7 +10,7 @@
 var $ = require('jquery');
 var Widgets = require('jupyter-js-widgets');
 
-    var WidgetManager = function(kernel, msgHandler) {
+    var WidgetManager = function(kernel, kernelname, msgHandler) {
```
jhpedemonte: `IKernel` (type of `kernel`) has a `name` property, but it always seems to come back as empty string in my testing.

jhpedemonte:
Notebooks (i.e. *test_widget_overlap*) which specify `python2` as kernel don't work correctly. We get these errors:
```
Error: Object 'ipython.widget' not found in registry
```
And I see this from the kernel gateway:
```
[IPKernelApp] ERROR | No such comm target registered: jupyter.widget.version
```

Running these notebooks in Notebook works just fine. Editing notebook to use `python3` instead makes them work correctly.

client/js/widget-manager.js
jhpedemonte: OK, so `kernel.name` is an empty string because we pass an empty string now when calling `startNewKernel`. Instead, I think we should do this:

```diff
diff --git a/client/js/dashboard.js b/client/js/dashboard.js
index fec0656..c88d37e 100644
--- a/client/js/dashboard.js
+++ b/client/js/dashboard.js
@@ -50,7 +50,7 @@ if (Element && !Element.prototype.matches) {
     var renderMime = _createRenderMime();
 
     // start a kernel
-    Kernel.start().then(function(kernel) {
+    Kernel.start(Config.kernelname).then(function(kernel) {
         // do some additional shimming
         _setKernelShims(kernel);
 
@@ -58,7 +58,7 @@ if (Element && !Element.prototype.matches) {
         _registerKernelErrorHandler(kernel);
 
         // initialize an ipywidgets manager
-        var widgetManager = new WidgetManager(kernel, Config.kernelname, _consumeMessage);
+        var widgetManager = new WidgetManager(kernel, _consumeMessage);
 
         // initialize Declarative Widgets library
         var widgetsReady = _initDeclWidgets();
diff --git a/client/js/kernel.js b/client/js/kernel.js
index 4444210..a15135a 100644
--- a/client/js/kernel.js
+++ b/client/js/kernel.js
@@ -18,14 +18,14 @@ var Services = require('jupyter-js-services');
 
     var _kernel;
 
-    function _startKernel() {
+    function _startKernel(kernelname) {
         var loc = window.location;
         var kernelUrl = loc.protocol + '//' + loc.host;
         var clientId = _uuid();
         var kernelOptions = {
             baseUrl: kernelUrl,
             wsUrl: kernelUrl.replace(/^http/, 'ws'),
-            name: '', // Set by API when making POST request to kernel gateway
+            name: kernelname,
             clientId: clientId,
             ajaxSettings: {
                 requestHeaders: {
diff --git a/client/js/widget-manager.js b/client/js/widget-manager.js
index 278b2e8..d3a717c 100644
--- a/client/js/widget-manager.js
+++ b/client/js/widget-manager.js
@@ -10,7 +10,7 @@
 var $ = require('jquery');
 var Widgets = require('jupyter-js-widgets');
 
-    var WidgetManager = function(kernel, kernelname, msgHandler) {
+    var WidgetManager = function(kernel, msgHandler) {
         //  Call the base class.
         Widgets.ManagerBase.call(this);
 
@@ -26,7 +26,7 @@ var Widgets = require('jupyter-js-widgets');
 
         // Validate the version requested by the backend -- required for ipywidgets
         // If kernel name is not defined, default to python3
-        kernelname = kernelname.toLowerCase();
+        var kernelname = kernel.name.toLowerCase();
         if (!kernelname || kernelname === 'python3' || kernelname === 'python2') {
             var validate = (function validate() {
                 this.validateVersion().then(function(valid) {
```

jameslmartin: I'm not sure if this is what we want to do here. The original bug was that the front end was hardcoding python3 into that field, `kernel.name`. @parente and I discussed that the back end should determine what type of kernel should be started, rather having the front end pass along the name.

However, we do need the kernel name for the widget manager. Thoughts?

parente: See below.

parente:
Help me disentangle.

> Notebooks (i.e. test_widget_overlap) which specify python2 as kernel don't work correctly. 

The kernel gateway dev Docker image needs to install ipywidgets against the python2 environment.

> 1. Kept getting validation error for ipywidgets code. We decided to expose the kernel name in the HTML template and only run the validation for python kernels.

Was this only for python2? If so, see above.  If it's for Toree too, are the necessary components for decl widgets installed in the kernel gateway container as well?

>  @parente and I discussed that the back end should determine what type of kernel should be started, rather having the front end pass along the name.

The frontend will never know the language of the dashboard unless its encoded in the HTML returned by the server. Even then, if that language is passed back to the server, it can't be trusted since the client can be manipulated.

Why exactly does the widget manager need to know the language?

> 2. Ran across the Missing property 'buffers' error. This happens when the Scala kernel craps out due to insufficient memory. Looks like the kernel gateway tries to restart it and sends back websocket messages with execution_state: "restarting". However, those don't have a buffer property, which the jupyter-js-services code expects (even if empty). Kernel gateway bug?

Kernel gateway is using the restart handler from notebook as-is. So doubtful. This is a separate issue about OOM handling, so let's put it aside for now.

client/js/widget-manager.js
```diff
@@ -24,17 +24,21 @@ var Widgets = require('jupyter-js-widgets');
...
-                }
-            }).catch(function(err) {
-                console.error('Could not cross validate the widget frontend and backend versions.', err);
-            });
-        }).bind(this);
-        validate();
+        // Validate the version requested by the backend -- required for ipywidgets
+        // If kernel name is not defined, default to python3
+        kernelname = kernelname.toLowerCase();
+        if (!kernelname || kernelname === 'python3' || kernelname === 'python2') {
```
parente: You'll also have to include just "python" here, but I really suspect the Python check won't be required once the other issues are resolved (e.g., properly installing the decl widgets scala jar, installing ipywidgets against python2 conda env) .

jameslmartin: > Why exactly does the widget manager need to know the language?
We need it here in this check to run the validation code.

parente: Right. I still think once Scala is setup properly, that special casing for Python will not be needed. I'll get with you and or @jhpedemonte today.

Dockerfile.kernel
```diff
@@ -1,7 +1,7 @@
 # Copyright (c) Jupyter Development Team.
 # Distributed under the terms of the Modified BSD License.
 
-FROM jupyter/pyspark-notebook:0017b56d93c9
+FROM jupyter/all-spark-notebook:0017b56d93c9
 
 RUN pip install jupyter_kernel_gateway==0.5.0
 # install ipywidgets 5.x
```
parente: This is where work needs to be done:

1. Declarative widgets and ipywidgets need to be installed against python2 environment.
2. Declarative widgets JAR needs to be installed against the Toree kernel.

jhpedemonte:
One reason we need the kernel name on the client side is because it does the call to `startNewKernel` for *jupyter-js-services*. The `Kernel` instance returned uses this passed in kernel name, not what is returned from backend (yay!). If we don't specify a kernel name to `startNewKernel`, then `kernel.name` is an empty string.

client/js/widget-manager.js
parente: Discussed with @jhpedemonte . For now, we're not going to bother setting up decl widgets for scala since the example is only hello world. Adding whatever config is necessary for a scala decl widgets example to work can come as part of #203.

parente: @jhpedemonte also clarified that the validation check does not actually inhibit the Toree kernel from working. It's just an ugly error. We agreed that switching it to a warning that explicitly says ala "Compatible version of ipywidgets is not available" is cleaner than hardcoding python checks and still makes sense with other kernels (yes, ipywidgets is not available for R, Toree, etc.)

parente:
> One reason we need the kernel name on the client side is because it does the call to startNewKernel for jupyter-js-services.

I'm fine including the kernel name in the template like you're doing as long as we put as many guards as we can in place to make sure that the kernel name determined on the *server-side* is the only one that's trusted and sent to the kernel gateway.

If there's an issue or question to open about how the name is used in jupyter-js-services, please go ahead and report it.

Dockerfile.kernel
parente: For first one, it's something like:

`RUN /opt/conda/envs/python2/bin/pip install jupyter_declarativewidgets` 

Check the path (should be in the docker-stacks README) and probably add a version too.

jhpedemonte:
Superseded by #236

PR #212, opened 2016-05-06
Update jupyter dependencies to latest

### LINKED ISSUE #141: Some dashboards not working in Safari

jhpedemonte:
Also contains fix for running scripts on Safari.

Fixes #141

### DIFF
--- a/client/less/notebook.less
+++ b/client/less/notebook.less
@@ -75,6 +75,10 @@
 
     .p-Widget {
         padding: 0.4em;
+
+        .jp-OutputArea-prompt {
+            display: none;
+        }
     }
 }
 
--- a/Dockerfile.kernel
+++ b/Dockerfile.kernel
@@ -4,8 +4,8 @@
 FROM jupyter/pyspark-notebook:0017b56d93c9
 
 RUN pip install jupyter_kernel_gateway==0.5.0
-# install ipywidgets 5.0
-RUN pip install 'ipywidgets==5.0'
+# install ipywidgets 5.x
+RUN pip install 'ipywidgets==5.1.3'
 
 # install Declarative Widgets python package
 # don't bother activating the extension, not needed outside notebook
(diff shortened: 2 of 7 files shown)


jhpedemonte:
Hmm, integration-test fails:

```
  1) dashboard with widgets "before all" hook:
     Uncaught Error: connect ECONNREFUSED 192.168.99.100:3000
      at Object.exports._errnoException (util.js:856:11)
      at exports._exceptionWithHostPort (util.js:879:20)
      at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1062:14)
```

Why would it refuse the connection?

jhpedemonte:
Forgot to include `commonmark` npm module, therefore dashboard server was failing to start as part of integration tests. Now everything passes. Good to go.

client/less/notebook.less
```diff
@@ -75,6 +75,10 @@
 
     .p-Widget {
         padding: 0.4em;
+
+        .jp-OutputArea-prompt {
+            display: none;
```
parente: So the prompts are still in there, just hidden?

parente:
`make clean examples kill build dev-install dev`. Every example notebook I tested worked, including the plotting one in Safari. I tested some additional ones without any major issues as well.

One new annoying behavior that I'm going to open as a follow-on issue. Every time I open a paper-dropdown, the entire browser page shifts to the right. It shifts back when I close it. It happens in every demo that uses the dropdown, and in both Chrome and Safari.

client/less/notebook.less
jhpedemonte: Right. That's written by the notebook's output area, so we can only hide it.

PR #213, opened 2016-05-06
[Enhancement] Split notebook-store into multiple modules

dalogsdon:
`notebook-store.js` contained several utility functions that should be individual modules. I've split it apart accordingly and cleaned up the code a bit.

This also includes the addition of a public dashboard link which is used in https://github.com/jupyter-incubator/dashboards_bundlers/pull/37.

### DIFF
--- a/app/notebook-fs.js
+++ b/app/notebook-fs.js
@@ -1,0 +1,122 @@
+/**
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+/**
+ * Augmented fs functions for notebook files.
+ */
+var appendExt = require('./append-ext');
+var config = require('./config');
+var fs = require('fs');
+var path = require('path');
+
+var DATA_DIR = config.get('NOTEBOOKS_DIR');
+var DB_EXT = config.get('DB_FILE_EXT');
+var INDEX_NB_NAME = config.get('DB_INDEX');
+
+function _ensureDataPath(nbpath) {
+    return nbpath.indexOf(DATA_DIR) === 0 ? nbpath : path.join(DATA_DIR, nbpath);
+}
+
+function _exists(nbpath) {
+    return new Promise(function(resolve, reject) {
+        nbpath = _ensureDataPath(nbpath);
+        fs.stat(nbpath, function(err, stats) {
+            if (err) {
+                if (err.code === 'ENOENT') { // file not found
+                    resolve(false);
+                } else {
+                    reject(err);
+                }
+            } else {
+                resolve(stats.isFile());
+            }
+        });
+    });
+}
+
+function _statError(err, nbpath) {
+    var e = new Error('Error getting notebook info: ' + nbpath + ' - ' + err.message);
+    if (err.code === 'ENOENT') {
+        e.status = 404;
+    }
+    return e;
+}
+
+function _statNbOrDir(nbpath) {
+    return new Promise(function(resolve, reject) {
+        // try to stat a notebook first, then fallback to original value
+        var extPath = appendExt(nbpath, DB_EXT);
+        fs.stat(extPath, function(err, stats) {
+            if (err && extPath === nbpath) {
+                reject(_statError(err, nbpath));
+            } else if (stats && stats.isFile()) {
+                stats.fullpath = extPath;
+                resolve(stats); // notebook exists
+            } else {
+                // path may be a directory (bundled dashboard) or other file
+                fs.stat(nbpath, function(err, stats) {
+                    if (err) {
+                        reject(_statError(err, nbpath));
+                    } else {
+                        stats.fullpath = nbpath;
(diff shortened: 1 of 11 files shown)


app/notebook-fs.js
```diff
@@ -0,0 +1,122 @@
...
+            return stats;
+        }
+    });
+}
+
+module.exports = {
+    /**
+     * Checks if the specified file exists
+     * @param  {String} nbpath - path to a notebook
+     * @return {Promise(Boolean)} resolved to true if it exists
```
jhpedemonte: -> `{Promise<Boolean>}`

app/notebook-fs.js
```diff
@@ -0,0 +1,122 @@
...
+     */
+    exists: _exists,
+    /**
+     * Runs `stat` on the specified path.
+     * @param {String} nbpath - Path to a notebook or directory.
+     *     If not relative to data directory, it will be made so.
+     * @return {Promise} resolved with the `stat` results. Will contain some custom properties:
```
jhpedemonte: -> `{Promise<Object>}`

app/notebook-store.js
```diff
@@ -2,133 +2,53 @@
...
-
+            var nbfile = stats.fullpath ||
+                path.join(DATA_DIR, appendExt(nbpath, DB_EXT));
             return new Promise(function(resolve, reject) {
-                var nbAbsPath = path.join(DATA_DIR, nbfile);
-                fs.readFile(nbAbsPath, 'utf8', function(err, rawData) {
+                fs.readFile(nbfile, ENCODING, function(err, rawData) {
                     if (err) {
-                        reject(new Error('Error loading notebook: ' + err.message));
+                        reject(err);
```
jhpedemonte: Why get rid of the descriptive error text?

app/notebook-store.js
```diff
@@ -2,133 +2,53 @@
...
-                        try {
-                            var nb = JSON.parse(rawData);
-                            // cache notebook for future reads -- use given `nbpath` since that
-                            // is path from request. later calls will look up using request path.
-                            store[nbpath] = nb; 
-                            resolve(nb);
-                        } catch(e) {
-                            reject(new Error('Error parsing notebook JSON'));
-                        }
+                        var nb = JSON.parse(rawData);
```
jhpedemonte: By removing the `try/catch`, any parse error will be caught by `new Promise` and result in rejection. I assume that's fine?

app/notebook-store.js
```diff
@@ -158,260 +78,26 @@ function get(nbpath, stats) {
...
     /**
-     * Uploads a notebook file
-     * @param {Request}  req - HTTP request object
-     * @param {Response} res - HTTP response object
-     * @param {Function} next - next function
+     * Removes the specified notebook from the cache
+     * @param {String} nbpath - path of notebook to remove from cache
      */
-    upload: upload
+    uncache: _uncache
```
jhpedemonte: I do not abide by this nomenclature.

routes/renderers.js
```diff
@@ -189,10 +177,6 @@ function _render(req, res, next, opts) {
             }
         })
         .catch(function failure(err) {
-            // If path on disk does not exist at all, return 404 not found.
-            if (err.code === 'ENOENT') {
-                err.status = 404;
-            }
```
jhpedemonte: Why remove this?

app/notebook-store.js
dalogsdon: Yes, we were already rejecting with a JSON parse error anyway, so I removed the try/catch to clean up the code a bit.

app/notebook-store.js
dalogsdon: Not sure... it looks like the old pattern is still being used below. I'll revert the error to the more descriptive.

app/notebook-fs.js
dalogsdon: It looks like they have defined [a syntax](https://github.com/jsdoc3/jsdoc/issues/1110) for documenting Promises but not implemented yet. In this example it would be `{Promise.<boolean>}`. I can update all the doc strings to this syntax.

routes/renderers.js
dalogsdon: See `notebook-fs.js` line 40. It is part of an effort to pull `fs` error code checking into the modules actually performing the filesystem operations. The HTTP error status can be set by the storage modules. This will help make it easier to swap the storage modules in the future if we want to switch from a filesystem store to some other storage solution.

app/notebook-store.js
dalogsdon: Provide an alternative. Perhaps `removeFromCache` or `invalidate`?

parente:
https://github.com/jupyter-incubator/dashboards_server/wiki/Server-API needs to be update with info about the response to POST.

dalogsdon:
Updated wiki page to mention dashboard link in response.

PR #63, opened 2016-02-04
WIP Deploy to Cloud Foundry

jtyberg:
This is a work in progress.  It includes changes required to build the application locally, then push successfully to Cloud Foundry.

PR #214, opened 2016-05-09
Example Cloud Foundry, Docker deployments

parente:
I have https://dashboards-server.mybluemix.net/ running temporarily to show the CF example works. I duplicated https://gist.github.com/jhpedemonte/0e156782a7039be0ecc6fa335390bed0 as a sibling of the CF example so that it doesn't look like we're favoring one of the other. Maybe this copy should become the canonical docker-compose example. It requires users to clone the dashboards_server repo just to get (bad) but is in a more discoverable / contributable (?) location (good).

Replaces #63


jhpedemonte:
While the CF deploy is mainly geared towards the dashboard server (which requires the kernel gateway), the Docker deploy also adds in the Notebook. For that reason, I'm wondering if it makes sense to be part of the `dashboards_server` git repo.

Maybe we should make a separate repo for these, since they bring together several different projects. This would also make it more palatable to have users clone, since it would only bring down a few files (similar to current gist), rather than the whole `dashboards_server` repo.

parente:
I wonder if we should repurpose https://github.com/jupyter-incubator/dashboards_tmpnb as dashboards_deploy, and make these two recipes + the tmpnb one all subfolders under that repo.

Thoughts?

parente:
Done. https://github.com/jupyter-incubator/dashboards_setup

PR #215, opened 2016-05-09
Ensure outgoing WS msgs are handled in order

### LINKED ISSUE #210: import pandas as pd, then pd.something yields pd not defined error
Looks very similar to the bug where certain cell executions were being skipped because of the way the DOM was laying them out. Just hit it again on 0.5.0. But this time, I don't think it's related to the size of the window or the DOM order.

Will provide websocket logs and details soon. Just noting it for now.

jhpedemonte:
Fixes #210 

@parente: Test this when you get a chance.


jhpedemonte:
Fixed unit tests. All tests now pass. Just need to test against actual failure case.

parente:
My go-to test case is fixed with your PR. Did you test other notebooks to make sure there are no surprises?

parente:
Nevermind. I beat it up and it looks fine. Let's :shipit:

PR #216, opened 2016-05-10
Fix relative path to config.json

parente:
Otherwise path to config is computed relative to cwd
which will likely be incorrect when run from an npm install

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -11,7 +11,7 @@ var debug = require('debug')('dashboard-proxy:config');
 var config = nconf.use('memory')
                   .argv()
                   .env()
-                  .file({ file: 'config.json', format: hjson });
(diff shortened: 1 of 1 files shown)


app/config.js
```diff
@@ -11,7 +11,7 @@ var debug = require('debug')('dashboard-proxy:config');
 var config = nconf.use('memory')
                   .argv()
                   .env()
-                  .file({ file: 'config.json', format: hjson });
+                  .file({ file: path.join(__dirname, '..', 'config.json'), format: hjson });
```
dalogsdon: A few lines below this there is the same path computed (`path.join(__dirname, '..', 'config.json')`). May want to save in a var.

parente: Will do.

Issue #196, opened 2016-04-20
New security review of 0.6.0


parente:
Update https://github.com/jupyter-incubator/dashboards/wiki/Deployed-Dashboard-Threat-Analysis based on 0.5.0 release.

parente:
#217 corrected a known security issue in 0.5.0. Moving this to 0.7.0 and a review of 0.6.0.

parente:
Added the new review to the wiki alongside the original. Opened issues for [planned countermeasures](https://github.com/jupyter-incubator/dashboards/wiki/Deployed-Dashboard-Threat-Analysis#planned-countermeasures).

PR #219, opened 2016-05-10
Filter 'code' prop on 'execute_input' replies to client

### LINKED ISSUE #217: Filter "code" property from WS msgs arriving to client
In execution replies that come back to the browser/client, there is a `code` property which shows the code that was executed on the kernel. We should filter this out in the proxy in order to make the code execution fully opaque from the client's view.

jhpedemonte:
Fixes #217


parente:
Code looks fine to me. I like that we have a consistent way to process any type of client or kernel originating message now. 

I'll put it through the paces a bit today before merging.

parente:
Everything I tried worked. I inspected the WS messages and saw no code. I looked at the message spec some more and found a few places where we might still have security holes. Will document those in #196  and we'll handle separately. For now, this one LGTM.

PR #226, opened 2016-05-11
[Issue 218] Make public link a customizable pattern

### LINKED ISSUE #218: Allow hostname, port, and protocol in PUBLIC_LINK
https://github.com/jupyter-incubator/dashboards_bundlers/pull/37 makes the response from the dashboard server the default place to get the link to a deployed dashboard instead of a config value in the bundler. Currently, the dashboard server uses a hardcoded URL as the base URL. The original logic in the bundler could compute certain properties such as the hostname, protocol, and port using the Host header. The same should be done here to make configurations more portable.

dalogsdon:
Allows {protocol}, {hostname}, and {port} placeholders in a PUBLIC_LINK_PATTERN config parameter

Upload API will build the public link based on this pattern and omit the link if the pattern is not set in the config

Resolves #218

### DIFF
--- a/app/public-link.js
+++ b/app/public-link.js
@@ -1,0 +1,26 @@
+/**
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+var config = require('./config');
+
+var PUBLIC_LINK_PATTERN = config.get('PUBLIC_LINK_PATTERN');
+var PROTOCOL = /\{protocol\}/g;
+var HOST = /\{host\}/g;
+var PORT = /\{port\}/g;
+
+/**
+ * Generates a base URL that can be used to build public links. Uses the pattern
+ * specified by the PUBLIC_LINK_PATTERN config parameter.
+ * @param  {Request} req - Express request used to get protocol and hostname
+ * @return {String} public base URL
+ * @return {undefined} if PUBLIC_LINK_PATTERN is not set
+ */
+module.exports = function(req) {
+    if (PUBLIC_LINK_PATTERN) {
+        return PUBLIC_LINK_PATTERN
+            .replace(PROTOCOL, req.protocol)
+            .replace(HOST, req.headers.host)
+            .replace(PORT, req.headers.host.split(':')[1] || '');
+    }
+};
--- a/config.json
+++ b/config.json
@@ -11,6 +11,14 @@
 
     // Mode in which the dashboard server express environment runs
     NODE_ENV: development
+    // Base URL pattern for building public link for an uploaded notebook.
+    // Supports placeholder strings (values come from request host header):
+    //   {protocol} - request protocol (HTTP or HTTPS)
+    //   {host} - value of the host header, (hostname[:port])
+    //   {port} - port extracted from the host header
+    // Examples using placeholders: {protocol}://{host}
+    //                              {protocol}://192.168.99.100:{port}
+    PUBLIC_LINK_PATTERN: http://127.0.0.1:3000
     // Interface on which the dashboard server listens
     IP: 127.0.0.1
     // Port on which the dashboard server listens
--- a/routes/auth-routes.js
+++ b/routes/auth-routes.js
@@ -7,21 +7,24 @@
  */
 var authToken = require('../app/auth-token');
 var config = require('../app/config');
+var link = require('../app/public-link');
 var upload = require('../app/notebook-upload');
 var router = require('express').Router();
 var urljoin = require('url-join');
 
-var GET_URL = urljoin(config.get('PUBLIC_LINK'), '/dashboards');
 var UPLOAD_MESSAGE = 'Notebook successfully uploaded';
 
 /* POST /notebooks/* - upload a dashboard notebook */
 router.post('/notebooks(/*)', authToken, upload, function(req, res) {
+    var publicLink = link(req);
     var resBody = {
-        link: urljoin(GET_URL, req.params[0]),
         message: UPLOAD_MESSAGE,
         status: 201,
         url: req.url
     };
+    if (publicLink) {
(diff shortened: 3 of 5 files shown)


app/public-link.js
```diff
@@ -0,0 +1,26 @@
...
+ * @param  {Request} req - Express request used to get protocol and hostname
+ * @return {String} public base URL
+ * @return {undefined} if PUBLIC_LINK_PATTERN is not set
+ */
+module.exports = function(req) {
+    if (PUBLIC_LINK_PATTERN) {
+        return PUBLIC_LINK_PATTERN
+            .replace(PROTOCOL, req.protocol)
+            .replace(HOSTNAME, req.hostname)
+            .replace(PORT, config.get('ACTIVE_PORT'));
```
parente: Is there a reason we're not pulling port from the Host header as well? The public facing port might not be known by the developer at deploy time (e.g., proxy defined).

bin/jupyter-dashboards-server
```diff
@@ -41,9 +42,12 @@ if (config.get('SSL_OPTIONS')) {
     }).listen(port, ip);
     debug('Server set to HTTPS mode');
 } else {
+    activePort = port;
     server = http.createServer(app).listen(port, ip);
     debug('Server set to HTTP mode');
 }
+// save the correct port for easy lookup in routes, etc.
+config.set('ACTIVE_PORT', activePort);
```
parente: This breaks in Docker containers among other places where the server port is mapped to an external port. The port has to be read from the request.

dalogsdon: Good point, we can parse the port out of host header.

app/public-link.js
dalogsdon: I was using the Express API which doesn't seem to include a port, but I can just manually parse the header.

dalogsdon:
I spoke with @jhpedemonte and he suggested the port could be treated special since it is optional in the host header. Does it make sense to have a {port} placeholder? I've updated the PR to support {protocol} and {host} placeholder. The host placeholder will get the value of the host header.

parente:
>  The host placeholder will get the value of the host header from the request.

If it's the whole host header with the port included, then support `{host}` and `{protocol}` is enough in the template.

dalogsdon:
> it's the whole host header with the port included

yes

routes/auth-routes.js
```diff
@@ -7,21 +7,24 @@
...
 var upload = require('../app/notebook-upload');
 var router = require('express').Router();
 var urljoin = require('url-join');
 
-var GET_URL = urljoin(config.get('PUBLIC_LINK'), '/dashboards');
 var UPLOAD_MESSAGE = 'Notebook successfully uploaded';
 
 /* POST /notebooks/* - upload a dashboard notebook */
 router.post('/notebooks(/*)', authToken, upload, function(req, res) {
+    var publicLink = link(req);
```
parente: Did you try this with the end-to-end docker compose? My gut tells me this will compute the wrong value because it's the bundler hitting the URL with its Host header (some internal private docker container name) instead of a user hitting it from the outside world with a browser.

I wonder if this needs to be computed once on first user visit to any user-facing dashboard server route and saved off. That's when the Host header will be correct for users visiting the server and that's the type of URL what we want to include in the POST response, something a user can be redirected to.

Yes, there's a problem with the case where the POST to notebooks happens before a user visits the server.

dalogsdon:
Discussed with @parente and decided the pattern will need to be set properly depending on the deployment. For Docker, the hostname will need to be set explicitly so it doesn't read some internal URL from the request from the bundler, for example. For our dev deployment, I added the docker-machine IP to the Makefile, so the pattern looks like `{protocol}://$$(docker-machine ip ...):{port}`. I reinstated the port placeholder to be able to do this (the host placeholder will still include the port).

parente:
I'll give this one more pass soon just to make sure it work and then merge, even if we have to iterate on it over time.

config.json
```diff
@@ -11,6 +11,12 @@
 
     // Mode in which the dashboard server express environment runs
     NODE_ENV: development
+    // Base URL pattern to use when generating public links
+    // Supported placeholder values:
+    //   {protocol} - request protocol (HTTP or HTTPS)
+    //   {host} - value of the host header, (hostname[:port])
+    //   {port} - port extracted from the host header
+    PUBLIC_LINK_PATTERN: "{protocol}://{host}"
```
parente: Hardcode this to `http://127.0.0.1` for consistency with the other defaults.

config.json
```diff
@@ -11,6 +11,12 @@
 
     // Mode in which the dashboard server express environment runs
     NODE_ENV: development
+    // Base URL pattern to use when generating public links
+    // Supported placeholder values:
```
parente: Needs more detail about where these values come from. Something like "The values are populated from the Host request header when a notebook is POSTed to the server."

routes/auth-routes.js
parente: We discussed offline. Leave this as is. Burden falls on the developer / deployer to set values properly. We can't infer everything in every situation.

parente:
Two more comments inline about config.json. Plus, the API wiki page needs a complete JSON response example in the upload section showing where the link is in the response (among all the other details of the response).

I'll merge after these are resolved.

dalogsdon:
Updated config.json and wiki page.

parente:
Comment in the config.json is still not clear that it's the Host header from the POST /notebooks request, not /dashboards or any other route. Merging and we'll update it if/when someone needs that detail.

dalogsdon:
It felt a little weird to do API documentation in the config file, since it would be another point to update if APIs change, so I kept it to just mentioning upload rather than POST /notebooks.

parente:
Agreed. That config.json is too overloaded. We'll need to write separate doc at some point.

PR #231, opened 2016-05-13
Cleanup same-name dashboards

### LINKED ISSUE #209: Upload with same name as existing should always "overwrite" existing
Two conditions currently exist:

1. Single-file and bundled dashboards are visibly identical in the UI.
2. Upload overwrites files.

It should follow that uploading a notebook with the same name as an existing bundled dashboard should delete the bundled dashboard, and vice versa.

For example, currently a `some-db.ipynb` dashboard can coexist with a `some-db` bundled dashboard since one is a file and the other is a directory. To reduce confusion about seemingly duplicate dashboards, we can enforce the proposed behavior above during upload.

jhpedemonte:
Fixes #209

### DIFF
--- a/app/notebook-upload.js
+++ b/app/notebook-upload.js
@@ -150,6 +150,7 @@ function _cleanPreviousDashboard(destination, destIsNotebook) {
             rimraf(destination, { disableGlob: true }, function(err) {
                 if (err) {
                     reject('Failed to remove existing bundled dashboard directory: ' + err.message);
+                    return;
                 }
                 resolve();
             });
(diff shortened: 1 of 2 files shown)


jhpedemonte:
@dalogsdon You know this code best, please review.

app/notebook-upload.js
```diff
@@ -133,6 +134,38 @@ function _writeZipFile(destination, buffer) {
...
+ */
+function _cleanPreviousDashboard(destination, destIsNotebook) {
+    return new Promise(function(resolve, reject) {
+        if (destIsNotebook) {
+            // destination is a notebook; ensure there isn't a directory by same name
+            rimraf(destination, { disableGlob: true }, function(err) {
+                if (err) {
+                    reject('Failed to remove existing bundled dashboard directory: ' + err.message);
+                }
+                resolve();
```
dalogsdon: Do you want to execute an unnecessary resolve after reject?

jhpedemonte: i think it's essentially a no-op, but i'll add a return to make it clearer.

dalogsdon: Yeah I was just going for code clarity.

dalogsdon:
Works for me.

PR #236, opened 2016-05-17
[Issue: 169] Frontend hardcodes python3

### LINKED ISSUE #169: Frontend hardcodes python3

jhpedemonte:
Fixes #169

### DIFF
--- a/Dockerfile.kernel
+++ b/Dockerfile.kernel
@@ -9,7 +9,13 @@ RUN pip install 'ipywidgets==5.1.3'
 
 # install Declarative Widgets python package
 # don't bother activating the extension, not needed outside notebook
-RUN pip install jupyter_declarativewidgets
+RUN pip install jupyter_declarativewidgets==0.5.2
+
+# also install above packages for python2
+USER root
+RUN /opt/conda/envs/python2/bin/pip install 'ipywidgets==5.1.3'
+RUN /opt/conda/envs/python2/bin/pip install jupyter_declarativewidgets==0.5.2
+USER jovyan
 
 # run kernel gateway, not notebook server
 CMD ["jupyter", "kernelgateway", "--KernelGatewayApp.ip=0.0.0.0"]
--- a/Dockerfile.kernel
+++ b/Dockerfile.kernel
@@ -3,18 +3,22 @@
 
 FROM jupyter/all-spark-notebook:0017b56d93c9
 
-RUN pip install jupyter_kernel_gateway==0.5.0
+ARG KERNEL_GATEWAY_VER=0.5.0
+ARG IPYWIDGETS_VER=5.1.3
+ARG DECLWIDGETS_VER=0.5.2
+
+RUN pip install jupyter_kernel_gateway==$KERNEL_GATEWAY_VER
 # install ipywidgets 5.x
-RUN pip install 'ipywidgets==5.1.3'
+RUN pip install "ipywidgets==$IPYWIDGETS_VER"
 
 # install Declarative Widgets python package
 # don't bother activating the extension, not needed outside notebook
-RUN pip install jupyter_declarativewidgets==0.5.2
+RUN pip install jupyter_declarativewidgets==$DECLWIDGETS_VER
 
 # also install above packages for python2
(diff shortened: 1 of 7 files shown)


Dockerfile.kernel
```diff
@@ -1,15 +1,21 @@
...
 # install Declarative Widgets python package
 # don't bother activating the extension, not needed outside notebook
-RUN pip install jupyter_declarativewidgets
+RUN pip install jupyter_declarativewidgets==0.5.2
+
+# also install above packages for python2
+USER root
+RUN /opt/conda/envs/python2/bin/pip install 'ipywidgets==5.1.3'
+RUN /opt/conda/envs/python2/bin/pip install jupyter_declarativewidgets==0.5.2
+USER jovyan
```
jhpedemonte: @parente Wouldn't install properly unless I was root. Is this the best way?

jhpedemonte:
OK, I think this is good to go. Runs scala test notebook (as well as all others) and passes integration tests.

Dockerfile.kernel
parente: No. Something's wrong if we can't run those as jovyan. I bet we need to activate the conda env properly. Like `bash -c "source activate python2 && <all the rest of the command lumped in>"`

parente:
Testing this AM.

routes/api.js
```diff
@@ -139,42 +133,62 @@ router.post('/kernels', bodyParser.json({ type: 'text/plain' }), function(req, r
...
+        // Store notebook path for later use
+        sessions[sessionId] = notebookPath;
+
+        nbstore.get(notebookPath)
+        .then(function success(notebook){
+            //debug(notebook.metadata);
+            if (notebook.metadata.kernelspec.name) {
+                var kernelName = notebook.metadata.kernelspec.name;
+                debug("Notebook kernel name found: " + kernelName);
+                if(kernelName === "apache_toree") {
```
parente: Why this mapping? If the kernelName in the notebook is "apache_toree", that should be how the kernel is requested from the KG as well. The notebook server definitely doesn't have any kind of special logic like this so we shouldn't either.

routes/api.js
```diff
@@ -139,42 +133,62 @@ router.post('/kernels', bodyParser.json({ type: 'text/plain' }), function(req, r
...
+                    kernelName = "scala";
+                }
+                req.body.name = kernelName;
+            } else {
+                // Default to Python 3
+                debug("Notebook kernel name not found, defaulting to Python 3");
+                req.body.name = "python3";
```
parente: Don't do this. The KG will be configured to pick the correct default kernel given a blank kernel name.

routes/api.js
parente: I see why. The `FROM` line in the `Dockerfile.kernel` is pointing to a build so old, that it predates the Toree kernel and is still using the ET-Spark kernel. Yikes!

So we definitely don't want the translation from apache_toree to scala here. It's making us compatible with the old ET kernel and not Apache Toree.

I suggest moving to 2d878db5cbff which is the last build to have notebook 4.1 and Toree (https://github.com/jupyter/docker-stacks/wiki/Docker-build-history).

etc/notebooks/test/test_scala_hello_world.ipynb
```diff
@@ -0,0 +1,67 @@
...
+      "layout": {
+       "col": 0,
+       "height": 4,
+       "row": 0,
+       "width": 4
+      }
+     }
+    }
+   },
+   "outputs": [
```
parente: Put a markdown cell somewhere in the dashboard. The hello world took so long to appear, I thought my dashboard server was busted. At least the markdown will appear instantly and give some sign of life. (Odd that the spinner bar didn't appear either ...)

jhpedemonte:
Update, @parente, please review.

parente:
LGTM. Thanks for finishing it off!

PR #241, opened 2016-05-19
Fix bad merge that broke travis build

parente:
Can't use Dockerfile ARGs on travis yet.
Using ENV for now instead.


parente:
Green on travis again. Merging it myself.

PR #244, opened 2016-05-23
Generate session secret if blank

parente:
Don't silently use a fixed, insecure default

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -77,7 +77,7 @@ if (key_file_location || cert_file_location) {
 }
 
 // Generate a session key if one is not specified. Log that we did it in case
-// the user accidentally left of an explict token.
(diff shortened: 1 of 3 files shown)


parente:
@jhpedemonte can you give this a quick look-see?

app/config.js
```diff
@@ -75,4 +76,12 @@ if (key_file_location || cert_file_location) {
     });
 }
 
+// Generate a session key if one is not specified. Log that we did it in case
+// the user accidentally left of an explict token.
```
jhpedemonte: of => off

jhpedemonte:
Looks good, only saw a minor misspelling. Fix that and merge, I say!

app/config.js
parente: All set.

Issue #227, opened 2016-05-12
New, extra padding in output area result


parente:
The cougars used to top-align. Now they don't.

![screen shot 2016-05-12 at 9 18 08 am](https://cloud.githubusercontent.com/assets/153745/15215842/87a8137c-1822-11e6-933c-2dc20a7c39bd.png)

Reason:

![screen shot 2016-05-12 at 9 18 13 am](https://cloud.githubusercontent.com/assets/153745/15215850/90355e32-1822-11e6-9c2f-e9c70a12b227.png)

Do we care?

dalogsdon:
Hmm Python vs Markdown images.Visually it would be nice but it could get really tedious fixing cases like  this, especially if the new Jupyter/Phosphor JS libs keep changing. I would say this could be something to look at in the future once things are more stable.

parente:
Yes. I'm not putting this into a milestone. Just wanted to note it in case one day we can do something about it easily.

PR #246, opened 2016-05-24
Fix extraneous cell padding

parente:
Dashboard preview in the 0.5.2 layout extension (which uses Gridstack) does not include any extra padding on cells:

screen shot 2016-05-24 at 9 21 48 am

Our read-only Gridstack-light clone here in dashboard server dynamically adds padding which causes full cells to overflow and misalign (see right edge of table):

screen shot 2016-05-24 at 9 21 06 am

This PR removes the extra padding. @jhpedemonte, @dalogsdon any reason why it should remain?

### DIFF
--- a/client/js/layout.js
+++ b/client/js/layout.js
@@ -65,9 +65,6 @@ var $ = require('jquery');
             sheet.insertRule('.dashboard-cell[data-layout-height="' + h + '"] { height: ' + height + '}', 0);
         }
 
-        // cell padding
-        sheet.insertRule('.dashboard-cell { padding: ' +
-            halfMargin + 'px ' + (halfMargin + 6) + 'px }', 0);
     }
 
     module.exports = {
--- a/etc/notebooks/test/test_padding.ipynb
+++ b/etc/notebooks/test/test_padding.ipynb
@@ -1,0 +1,157 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "urth": {
+     "dashboard": {
+      "layout": {
+       "col": 6,
+       "height": 5,
+       "row": 0,
+       "width": 6
+      }
+     }
+    }
+   },
+   "source": [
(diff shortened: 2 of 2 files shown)


dalogsdon:
The padding is based on the cell margin, so it could have something to do with the space between cells. However, in this case it looks wrong. Make sure removing it preserves the apparent space between cells in a grid.

parente:
> Make sure removing it preserves the apparent space between cells in a grid.

I'll dig a bit more into what is providing the space in the gridstack case today vs what dashboard server is doing.

parente:
Added a test_padding notebook. Here it is in notebook server with jupyter_dashboards 0.5.2:

screen shot 2016-05-26 at 9 28 28 am

Here it is in dashboard server WITH the fix in this PR:

screen shot 2016-05-26 at 9 29 14 am

Here it is in dashboard server WITHOUT the fix in this PR:

screen shot 2016-05-26 at 9 36 35 am

Note the excessive left-right padding without the fix.

dalogsdon:
I am not sure completely removing the padding is the correct solution. In the notebook, we add margin to the `inner-cell` container to visually add space between each cell (seen on hover when authoring dashboard). The padding in dashboard server was an attempt to mimic that so the spacing is identical, so perhaps it just needed to be modified rather than removed.

In the colors example above (with @parente's fix), I think it just happens to look the same between the notebook and the dashboard server because the phosphor widgets have their own padding that happens to match the cell margin we were adding in the notebook. This exposes another issue since the notebook is not yet using the new JS libs that use Phosphor widgets. Perhaps this was the original issue in the first place instead of our padding being wrong.

I made another example notebook using Markdown cells containing images and observed a spacing difference (with @parente's fix):

Notebook:
image

Dashboard server:
image

parente:
1. That's a scary set of images.
2. Let's pair on seeing if modification rather than removal corrects all cases. There is #227 about the phosphor differences which I think are separate.

PR #252, opened 2016-05-26
Another attempt at fixing cell padding

parente:
* Keep padding, but change values
* Handle rendered_html case

For #246


dalogsdon:
Looks good, the padding test and my test case using Markdown look closer between the notebook and dashboard server.

PR #255, opened 2016-06-01
Override WS logging with non-buffering version

jhpedemonte:
The websocket lib has a buffering logger which doesn't print out ws info in time with other logging messages. This PR overrides with the standard `debug` logger we use elsewhere.

To enable ws logging, add `websocket:*` to `start-logging` script in *package.json*.


jhpedemonte:
Travis failed due to fetch error creating docker image. I ran `make test integration-test` locally and all passed.

PR #256, opened 2016-06-03
Update to use new metadata spec

### LINKED ISSUE #224: Remove urth, support v1 metadata spec
Wait for https://github.com/jupyter-incubator/dashboards/issues/243 to get implemented, at least in a branch. Then:

* [ ] Remove all urth references in CSS classes and whatnot
* [ ] Support rendering of v1 metadata spec notebooks (https://github.com/jupyter-incubator/dashboards/wiki/Dashboard-Metadata-and-Rendering is finalized)
* [ ] Decide if/how we continue to support rendering of v0 spec notebooks

DO NOT add any new user-facing features (e.g., NO multi-layout support). This is a cleanup and switcheroo only.

jhpedemonte:
Also update example/test notebooks

Fixes #224 

cc @dalogsdon


app/handlebars-helpers.js
```diff
@@ -4,29 +4,22 @@
...
-            return {
-                layout: 'grid',
-                cellMargin: config.get('DB_CELL_MARGIN'),
-                defaultCellHeight: config.get('DB_DEFAULT_CELL_HEIGHT'),
-                maxColumns: config.get('DB_MAX_COLUMNS')
-            };
-        } else {
-            return {};
-        }
-    },
```
jhpedemonte: @dalogsdon: Do we need to have defaults for these values? Or will all created dashboards have it? What about notebooks with no layout?

If we don't need these any more, then we can remove from **config.json**, right?

dalogsdon: The dashboard server will generate these in every notebook. The question is do we want to provide defaults if they are missing for any reason (like if a user deleted them) or always require them in the notebook file? If we always require them, we can remove defaults from `config.json` here.

routes/renderers.js
```diff
@@ -88,25 +89,20 @@ function _renderDashboard(req, res, next, opts) {
...
             // get dashboard layout from notebook-level dashboard metadata
-            var dashboardLayout = (notebook.metadata &&
-                                  notebook.metadata.urth &&
-                                  notebook.metadata.urth.dashboard &&
-                                  notebook.metadata.urth.dashboard.layout) ||
-                                  NO_LAYOUT;
+            var activeView = getObject(notebook, 'metadata.extensions.jupyter_dashboards.activeView') ||
+                             NO_LAYOUT;
 
             // backwards compatibility - old grid layouts don't have layout set
```
dalogsdon: Are we providing backwards compatibility in dashboards server?

cc @parente

dalogsdon:
It looks like some notebook checkpoint files were checked in. See 2 inline comments above. Other than that, works for me and looks good.

routes/renderers.js
jhpedemonte: We can delete this backwards-compat layer.

app/handlebars-helpers.js
jhpedemonte: I think bulletproofing in this way should be handled in a separate issue. Right now, the code assumes that the dashboard was correctly generated by the dashboard extension.

jhpedemonte:
Updated based on comments. Please re-review when you get a chance, @dalogsdon.

Issue #259, opened 2016-06-17
Change to new location of jupyter-js-*


parente:
jupyter-js-ui and jupyter-js-notebook are now in one jupyterlab package. Need to update our deps and webpack config to pull in the pieces we need.

* https://gitter.im/jupyter/jupyterlab?at=5764031391d56ac9155c4ebf (and down)
* https://www.npmjs.com/package/jupyterlab
* https://github.com/jupyter/jupyterlab/blob/master/tutorial/repo.md

parente:
Note that jupyter-js-services and jupyter-js-utils are still separate.

parente:
jupyter-js-notebook is now in https://github.com/jupyter-attic.

PR #262, opened 2016-07-18
Fix bad text, last widget in tax_demo_grid

parente:
Spotted as part of #259


jhpedemonte:
Perfect. Works for me.

jhpedemonte:
Tests failing due to jquery-ui changes, which will be fixed later on. Merging this.

PR #263, opened 2016-07-20
Adds ability to delete dashboards (notebooks)

### LINKED ISSUE #250: A way to delete notebooks/dashboards
The server has an API for posting notebooks. There's no API for deleting them (e.g., if one is accidentally posted, if it's old and should be removed.) 

There's also no UI for using such an API anywhere yet. IMHO, we shouldn't grow an admin UI here since the dashboard server is more likely to be used as part of a bigger solution where that UI will be implemented elsewhere (e.g., Jupyter Hub). 

A simple CLI, though, would also solve #208 ...

jhpedemonte:
Adds integration test and fixes some others

Fixes #250

### DIFF
--- a/routes/auth-routes.js
+++ b/routes/auth-routes.js
@@ -29,6 +29,7 @@ router.post('/notebooks(/*)', authToken, upload, function(req, res) {
     res.status(201).json(resBody);
 });
 
+/* DELETE /notebooks/* - delete a dashboard notebook */
 router.delete('/notebooks(/*)', authToken, nbdelete);
 
 module.exports = router;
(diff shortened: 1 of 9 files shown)


routes/auth-routes.js
```diff
@@ -28,4 +29,6 @@ router.post('/notebooks(/*)', authToken, upload, function(req, res) {
     res.status(201).json(resBody);
 });
 
+router.delete('/notebooks(/*)', authToken, nbdelete);
```
parente: Maybe a short comment like the one on post() for new folks who might read it as taking some delete action immediately when called.

routes/presentation.js
```diff
@@ -6,7 +6,7 @@
  * Presentation routes. Only renders notebook in root directory.
  */
 var config = require('../app/config');
-var renderers = require('./renderers');
+var renderers = require('../app/renderers');
```
parente: This is a change I can get behind! 🇺🇸

PR #265, opened 2016-07-21
Adds REST API allow clearing of all or some cached dashboards

### LINKED ISSUE #208: A way to break notebook cache
Notebooks are cached in memory after the first time they are loaded for viewing as dashboards. The cache is only invalidated if a new version of the notebook is POSTed to /_api/notebooks. However, notebooks might be deployed by simply copying them into place on disk, requiring another way to invalidate cache.

Ideas:

* URL parameter `no_cache=true` forces a reload of the notebook from disk (requires user to know to use it or admin to use it once after overwriting notebooks on disk)
* DELETE /_api/cache resets the entire cache (requires admin to make the call)
* CLI that makes the DELETE call for convenience (icing)
* Improvement to the caching logic to know when a file has changed on disk (hard)

jhpedemonte:
Also adds missing delete test cases using auth token.

Fixes #208

* [x] REST API to clear cache
* [x] Command line script to call new REST API
* [x] Add dashboard upload handling to script
* [x] Add dashboard deletion handling to script
* [x] Package script with npm package and update name
* [x] Update API wiki page ([changes](https://github.com/jupyter-incubator/dashboards_server/wiki/Server-API/_compare/c4d210c3ab55f262ad2bbf7151465c838252b60d...e02c35bbb3ace96d4f2da03fbe94145fca827d5c))


parente:
Does it make sense to rename the clear cache script to jupyter-dashboards-admin and support other ops like deletion, no just cache clearing?

Does the script get bundled into the npm package? There's a make target for building a release without sending it to npm I think.

The API wiki page should get updated.

jhpedemonte:
As in deletion of a notebook/dashboard? Sure, we can add that. Could add upload as well.

We did remember the API wiki page, just haven't done that yet. And we'll take a look at npm package and naming.

jhpedemonte:
There's just the `npm publish`. But I was able to verify using `npm pack` that everything in `/bin` is included.

bin/jupyter-dashboard-admin
```diff
@@ -0,0 +1,139 @@
...
+    },
+    'auth-token': {
+        describe: 'authorization token for admin privelages, defaults to reading from config file',
+        type: 'string'
+    }
+};
+
+// command to clear the dashboard server cache
+var clear_cache_command = {
+    command: 'clear-cache [options]',
```
parente: The common options don't appear when passing `--help` for me so it's not clear how to pass the host or token without cracking the source. Can we fix this?

bin/jupyter-dashboard-admin
```diff
@@ -0,0 +1,139 @@
...
+var path = require('path');
+var request = require('request');
+var urljoin = require('url-join');
+var yargs = require('yargs');
+
+// Config defaults are in an HJSON file in the root of the source tree
+var defaultConfig = path.join(__dirname, '..', 'config.json');
+var config = nconf.file({ file: defaultConfig, format: hjson });
+
+var DEFAULT_HOST = 'http://localhost:3000';
```
parente: Can the default URL come from the config module so that the two are never out of sync?

parente:
Few problems using the CLI for uploads:

1. Uploaded notebook does not appear on disk or in the dashboard listing. Not sure where it went.

```
cd jupyter_dashboards/bin
./jupyter-dashboard-admin upload ../etc/notebooks/test/test_layout.ipynb test/test_layout2.ipynb
Dashboard successfully uploaded!
ls -l ../data/test
# test_layout2.ipynb is not here
```

2. File upload completely replaces an existing directory of the same name. This should probably give a warning or error in the CLI vs allowing the overwrite like we do from the bundler. (Why: the human is giving the path here and can mess it up vs the bundler code in the original case.)

```
cd jupyter_dashboards/bin
./jupyter-dashboard-admin upload ../etc/notebooks/test/test_layout.ipynb test
Dashboard successfully uploaded!
ls -l ../data
# test/ folder is completely deleted, only test notebook remains
```

3. Notebook never uploaded to non-existent subpath. Subpath gets created, but notebook / bundle is not written to disk.

```
cd jupyter_dashboards/bin
./jupyter-dashboard-admin upload ../etc/notebooks/test/test_layout.ipynb new_folder/test_layout.ipynb
Dashboard successfully uploaded!
ls -l ../data/new_folder
# nothing in the folder
```

jhpedemonte:
> File upload completely replaces an existing directory of the same name.

This is the standard behavior of the REST API for `/upload`.

parente:
> This is the standard behavior of the REST API for /upload.

Yes. Understood. I was poorly explaining that the behavior is correct for the bundler type action where a user just clicks "Deploy to dashboard server" and the right thing happens. But from a CLI, where a human has to type a file name, we might want the CLI to at least give a warning if something with that name already exists because the human made a typo or some such.

jhpedemonte:
Without changing the current REST API, we could warn the user that a dashboard by that name already exists (whether an individual file or bundled) and ask if user wants to continue. Does that seem fine to you? Or did you have something else in mind?

parente:
> Without changing the current REST API, we could warn the user that a dashboard by that name already exists (whether an individual file or bundled) and ask if user wants to continue.

Yea. That's what I was thinking. Something simple in the CLI by doing a GET first to check existence. I definitely don't want to change the API.

bin/jupyter-dashboard-admin
jhpedemonte: First, don't want to include **config.js**, since that handles its own `nconf` instance and does other things which we don't want to do in this CLI app.

I could read in and parse `PUBLIC_LINK_PATTERN`, replacing `{host}` & `{port}` with `IP` & `PORT` config options. Default to `http` for `{protocol}`.

jhpedemonte:
> Uploaded notebook does not appear on disk or in the dashboard listing

This is because the destination name is listed with the `.ipynb` extension. It's a bug that it says success when nothing happened. But I also need to make it more evident in the help that the destination should be a dashboard name, no extension.

jhpedemonte:
OK, I think I took care of all of the issues found by Pete. @parente and @jameslmartin, please take a look at latest changes.

Also, I [updated the wiki](https://github.com/jupyter-incubator/dashboards_server/wiki/Server-API/_compare/c4d210c3ab55f262ad2bbf7151465c838252b60d...e02c35bbb3ace96d4f2da03fbe94145fca827d5c).

jhpedemonte:
Checks should pass once PR #275 is merged.

parente:
I'll give this a shot tomorrow.

parente:
Tested locally. Everything looks good now.

@jhpedemonte You do the merge since you're also working on #275. I don't know if you want to fix that first or get this in or ...

jhpedemonte:
Reran Travis tests since PR #278 went in. Now green so merging.

PR #266, opened 2016-07-22
Adds auth and/or extra headers to downstream requests

### LINKED ISSUE #264: Support basic auth and/or custom headers when proxying to a kernel gateway
A common pattern for securing access to one or more kernel gateways is to use a proxy server. For instance, an nginx proxy might have routes to multiple instances and secure them with different username/password combos. (See https://github.com/jupyter/kernel_gateway_demos/pull/24) As another example, the proxy might require auth headers other than the one supported by the kernel gateway (e.g., X-Instance).

To work with these setups, the dashboard server needs to support setting configurable headers when proxying HTTP or websocket connections. This could be implemented in a couple of ways. (Supporting one way is enough).

- Reading an object from the config with header key/value pairs
- Supporting a module system like we did with auth to plug-in in an arbitrary JS function that returns an object with header key/value pairs on each request

The former is probably enough for now, unless we thinking dynamic headers should be supported right off the bat.

jhpedemonte:
... pulled from config.json

Fixes #264

### DIFF
--- a/app/ws-rewriter.js
+++ b/app/ws-rewriter.js
@@ -30,6 +30,8 @@ var FRAGMENTATION_THRESHOLD = 0x3200000;
  * @param  {String} args.host - kernel gateway host/domain
  * @param  {String} args.basePath - kernel gateway base URL
  * @param  {Object} args.headers - kernel gateway headers
+ * @param  {Object} args.requestOptions - additional options passed to http(s).request() when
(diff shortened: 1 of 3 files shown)


jhpedemonte:
Tested against *node_client_example* from **kernel_gateway_demos**.  Auth worked and we inspected HTTP packets to ensure that headers are properly getting passed as well.

PR #267, opened 2016-07-25
Adds example Scala notebook with declarative widgets for testing purposes

jameslmartin:
Completes last task for #203


parente:
Works for me. There's a long pause with no indication of things happening as the kernel is firing up. I'll open a separate issue. Maybe we can get the progress bar to appear immediately on page load while waiting for the kernel.

PR #270, opened 2016-07-26
Busy / error improvements

### LINKED ISSUE #269: Long pause with no indicator while Toree kernel loads
The blue busy bar doesn't appear until the first busy message from the kernel. This leaves the user wondering what's happening while the kernel is spinning up.

Perhaps we should show the progress bar either before requesting a new kernel or immediately after the server responds that the kernel has been started (whichever fixes the issue and doesn't show a spinner indefinitely when there's an error.)

parente:
Show busy bar during kernel start. Show error indicator when kernel fails to start.

Fixes #269

### DIFF
--- a/client/js/dashboard.js
+++ b/client/js/dashboard.js
@@ -64,7 +64,7 @@ if (Element && !Element.prototype.matches) {
         var widgetsReady = _initDeclWidgets();
 
         // ensure client-side widgets are ready before executing code
-        widgetsReady.then(function() {
+        return widgetsReady.then(function() {
             // setup and execute code cells
             _getCodeCells().each(function() {
                 var $cell = $(this);
@@ -103,12 +103,9 @@ if (Element && !Element.prototype.matches) {
                 // widget *subarea* with its output areas and DOM container
                 widgetManager.trackPending(kernelFuture, $widgetSubArea.get(0), model);
             });
-        })
-        .catch(function(err) {
-            console.error(err);
-            ErrorIndicator.show();
         });
-    }).catch(function(e) {
+    }).catch(function(err) {
+        console.error(err);
         ErrorIndicator.show();
     });
 
--- a/client/js/dashboard.js
+++ b/client/js/dashboard.js
@@ -108,6 +108,8 @@ if (Element && !Element.prototype.matches) {
             console.error(err);
             ErrorIndicator.show();
         });
+    }).catch(function(e) {
+        ErrorIndicator.show();
     });
 
     // shim Jupyter Notebook objects for backwards compatibility
--- a/client/js/kernel.js
+++ b/client/js/kernel.js
@@ -54,7 +54,7 @@ var Services = require('jupyter-js-services');
                 kernel.statusChanged.connect(function(_kernel, status) {
                     clearTimeout(debounced);
                     debounced = setTimeout(function() {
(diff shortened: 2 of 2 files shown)


client/js/dashboard.js
```diff
@@ -108,6 +108,8 @@ if (Element && !Element.prototype.matches) {
             console.error(err);
             ErrorIndicator.show();
         });
+    }).catch(function(e) {
+        ErrorIndicator.show();
```
jhpedemonte: I think we can make this cleaner:

1. Return the promise from `widgetsReady.then()` (line 67).
2. Remove the catch for that line, lines 107-110.
3. Add `console.error(err)` to your new code.

This way, there's only one error handler rather than 2.

client/js/kernel.js
```diff
@@ -43,17 +54,14 @@ var Services = require('jupyter-js-services');
                 kernel.statusChanged.connect(function(_kernel, status) {
                     clearTimeout(debounced);
                     debounced = setTimeout(function() {
-                        var isBusy = status === 'busy';
-                        $('.busy-indicator')
-                            .toggleClass('show', isBusy)
-                            // Prevent progress animation when hidden by removing 'active' class.
-                            .find('.progress-bar')
-                                .toggleClass('active', isBusy);
+                        var isBusy = status === Services.KernelStatus.Busy;
```
jhpedemonte: I thought `Services.KernelStatus` doesn't exist anymore in latest **jupyter-js-services**. I had previously changed all references to use string values instead.

parente: Errr. That's a bad rebase. I'll fix.

client/js/dashboard.js
parente: What about the console.error happening down in kernel.js? Was thinking it's kinda weird that we do some logging in one place and other logging/error handling in another? Should that move up too?

jhpedemonte: Oh, yes, I see. So a unified error handler could get errors from (1) `Kernel.start` (in which case it has been logged) ; or (2) `widgetsReady` (where it has not been logged).

I think it makes the most sense to remove logging from `Kernel.start`, since it is also throwing the error. It should probably just create a new error with the text `Failed to create kernel`, wrapping the incoming error. Make sense?

parente:
@jhpedemonte I think I addressed your comments in the way you were saying. I tested the error handling by killing the KG and watching for the indicator. I tried to test against the scala widgets notebook, but it seems to have stopped working since the PR this morning. (I'm guessing it's the external declarativewidgets.jar that's not getting pulled now for some external reason ...)

client/js/kernel.js
```diff
@@ -44,18 +55,14 @@ var Services = require('jupyter-js-services');
...
+                        showBusyIndicator(isBusy);
                     }, 500);
                 });
                 return kernel;
             })
             .catch(function(e) {
-                console.error('failed to create kernel', e);
-                throw e;
+                showBusyIndicator(false);
+                throw new Error('Failed to create kernel');
```
jhpedemonte: I would change this to:

`throw new Error('Failed to create kernel: ' + e.message)`

This will provide more information.

jhpedemonte:
Just one minor nit. Otherwise, it's good to merge.

parente:
Fixed nit. Merge away when green.

PR #272, opened 2016-07-26
Allows read only NOTEBOOKS_DIR v2

bodri5:
Revised version of #268

### DIFF
--- a/app/config.js
+++ b/app/config.js
@@ -45,11 +45,9 @@ if(!nbDir) {
     nbDir = path.resolve(nbDir);
     // Create directory if it does not exist
     try {
-        fs.mkdirSync(nbDir);
+      fs.statSync(nbDir);
     } catch(e) {
-        if(e.code !== 'EEXIST') {
-            throw e;
-        }
+      fs.mkdirSync(nbDir);
     }
     config.set('NOTEBOOKS_DIR', nbDir);
 }
--- a/app/config.js
+++ b/app/config.js
@@ -45,11 +45,15 @@ if(!nbDir) {
     nbDir = path.resolve(nbDir);
     // Create directory if it does not exist
     try {
-        fs.statSync(nbDir);
+        var st = fs.statSync(nbDir);
     } catch(e) {
         fs.mkdirSync(nbDir);
     }
-    config.set('NOTEBOOKS_DIR', nbDir);
+    if (st.isDirectory()) {
+        config.set('NOTEBOOKS_DIR', nbDir);
+    } else {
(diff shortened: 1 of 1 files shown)


jhpedemonte:
Assuming we still want to attempt to create the directory if it doesn't exist, then the patch needs some work. I'll add comments inline.

app/config.js
```diff
@@ -45,11 +45,9 @@ if(!nbDir) {
     nbDir = path.resolve(nbDir);
     // Create directory if it does not exist
     try {
-        fs.mkdirSync(nbDir);
+      fs.statSync(nbDir);
```
jhpedemonte: Keep same indentation.

app/config.js
```diff
@@ -45,11 +45,9 @@ if(!nbDir) {
...
     } catch(e) {
-        if(e.code !== 'EEXIST') {
-            throw e;
-        }
+      fs.mkdirSync(nbDir);
```
jhpedemonte: I think it makes more sense to do something like this:

```javascript
if (e.code === 'ENOENT') {
    // dir doesn't exist; try to create it
    try {
        fs.mkdirSync(nbDir);
    } catch(e) {
        console.error('Failed to create required directory "' + nbDir + '": ' + e.message);
        throw e;
    }
} else {
    throw e;
}        
```

bodri5:
I fixed the indentations.

I do not quite understand @jhpedemonte's other comment. My code works the following
way: first stat the path with fs.statSync. If it does not exist, then  fs.statSync throws an exception,
and only then we try to create the directory with fs.mkdirSync. As we are already sure, that it does
not exists, any exception from mkdir can abort the program unconditionally.

jhpedemonte:
@bodri5: My suggested code change was meant to address two things:

1. I prefer to be explicit about what the code is doing. `fs.statSync` can throw many exceptions. The only one we want to handle is `ENOENT` (does not exist). All other exceptions should abort the program (hence the additional `throw e` in the `else` case).
2. `any exception from mkdir can abort the program unconditionally` -> This is true and also handled in my code sample above. I just wanted to explicitly write out an error string that is more understandable to the user (the `console.error` line).

Make sense?

bodri5:
@jhpedemonte: Thank you for the explanation. Modified the patch according to your suggestions.

jhpedemonte:
Travis was having issues, but now this is passing. I tested locally as well, pointing `NOTEBOOKS_DIR` at legitimate directories, a file and dir to which I didn't have permissions. All worked as expected. Merging.

Thanks @bodri5

PR #275, opened 2016-07-28
Adds missing loader for webpack

jhpedemonte:
Needed to properly build Jupyterlab 0.1.2

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -42,6 +42,7 @@ var webpackConfig = {
         loaders: [
             { test: /\.css$/, loader: 'style-loader!css-loader' },
             { test: /\.json$/, loader: 'json-loader' },
+            { test: /\.html$/, loader: 'file' },
             // jquery-ui loads some images
             { test: /\.(jpg|png|gif)$/, loader: "file" },
             // required to load font-awesome


jhpedemonte:
Making this WIP. They changed rendermime again. Yay!

jhpedemonte:
Everything is working with this patch except for matplotlib (*test_layout* & *test_plotting_libs*). Looks like messages trying to use matplotlib are being handled before `register_target('matplotlib', ...)` gets invoked.

jhpedemonte:
@blink1073 @jasongrout:

 I'm getting `Error: Object 'matplotlib' not found in registry` when building against Jupyterlab 0.1.2 (everything works fine with v0.1.1).  I never see call to `register_target('matplotlib')` before this error. Looks like perhaps a timing issue.

I notice that in *src/notebook/output-area/widget.ts*, the call to `rendermime.render` is not async. Could this be causing issue? Is the matplotlib JS not added to the DOM before handling incoming matplotlib comm messages?

blink1073:
This is using `%matplotlib notebook`?  I would not have expected that to work at all yet.

jhpedemonte:
It's been working up until now. We did have to add some [shim code](https://github.com/jupyter-incubator/dashboards_server/blob/master/client/js/widget-manager.js#L212) to make it all work, though.

blink1073:
I confess I haven't dealt much with the ipywidgets end of things.  Jason has already left for week-long holiday and I am leaving tomorrow for a week as well.  I'll see if I can figure something out this afternoon.

blink1073:
Also, we changed rendermime again in https://github.com/jupyter/jupyterlab/pull/560 (not released yet).

blink1073:
FYI @jasongrout, here's another example where we're not properly following semver, since this library depends on jupyterlab, we should be making minor releases, not patch releases.

jhpedemonte:
OK. I think I'll just pin Jupyterlab to v0.1.1 for now, so we can continue working. And I'll continue investigation of v0.1.2+.

blink1073:
Sounds good.

jhpedemonte:
Got it working. Looks like the fixes in jupyter/jupyterlab#560 (specifically the commit **Make rendering synchronous and cleanup**) fixes the issues I was seeing.  I updated this PR to work against v0.1.3 of Jupyterlab (not yet release).

blink1073:
Great, glad you got it sorted.  Hopefully we don't have to change this part of the code base too much anymore.  ;)

jhpedemonte:
Updated to use Jupyterlab 0.2.0.  Ran into codemirror/CodeMirror#4190, which breaks our build. Got past it by temporarily building with CodeMirror 5.17.0.

Plotting libs are broken once again (yay!) but padding seems to be fine now.

PR #291, opened 2016-09-21
Support BASE_URL

danielfrg:
This is a possible fix for https://github.com/jupyter-incubator/dashboards_server/issues/260.

I am not a node dev so a probably some things could be improved, let me know and I can make changes. I have tested it a little bit and the only think that its not working is the font awesome assets are still trying to be located from `/components/...` but everything else looks fine.

Issue #260, opened 2016-06-28
Serve dashboards with a prefix


ch41rmn:
Hi,

I recently tried to collect the several web services under one Apache server, one of them is the dashboards server. Each of these services run on a different localhost port, and Apache acts as the reverse-proxy to direct different URLs to different ports based on the prefix.

e.g.
`my.domain.com/jupyter --> http://127.0.0.1:9500/jupyter`, jupyter hub serving on `localhost:9500/jupyter`
`my.domain.com/jenkins --> http://127.0.0.1:8080/jenkins`, jenkins serving on `localhost:8080/jenkins`
...`

I cannot find a way to specify a prefix for the dashboards server. I can only get it to serve on `http://host:port`, whereas I would like to be able to serve at `http://host:port/prefix`.

My initial dig into this points to some changes required in the templates, particularly the way `/css` and `/components` are linked in the templates.

Is this an easy thing to implement? I am not very familiar with node/express, but will be happy to have a go at a pull request.

Cheers,

parente:
@ch41rmn We'd love a PR if you're willing to tackle this.

You are correct that the dashboard server is not currently relocatable under different paths. I'm pretty certain there is no major impediment preventing it, just a matter of finding all the spots that might break. The ones that come to mind are:

* The templates, as you pointed out. These should use relative path references wherever possible. If they can't they'll need to read a base URL config option that has to be defined.
* The express router. The root path needs to be made configurable via a base URL option. Hopefully all of the changes should be in app.js, either by adding the base URL value to all of the `app.use()` calls or by defining a new root router and tacking all of the exiting paths onto it. (ref: http://bulkan-evcimen.com/using_express_router_instead_of_express_namespace.html)
* Any places that generate links, like https://github.com/jupyter-incubator/dashboards_server/blob/792dc792cfa5358baede419b8918dff30dd1db16/routes/auth-routes.js#L26 (though that one is already driven by configuration, so might not need any work.)

@jhpedemonte, @dalogsdon can you think of other spots that would need updating?

dalogsdon:
I think those mentioned by @parente are the main points.

* For templates (mainly the listing view but possibly others) we are using Handlebars. You will probably need to pass the prefix when rendering. In the template, if a variable isn't set, it will resolve to an empty string.
* For the router, I'd vote for a root router. That way it's only in once place and you don't have to think about the prefix in the child routes.
[...]

PR #292, opened 2016-09-23
Support a 'prefix' url under which to run server (i.e. for proxies)

jhpedemonte:
Ref #260

Based on PR #291 by @danielfrg

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,16 +62,16 @@ kill: ## Kill all running docker containers
 dev-install: ## Install all dev deps on localhost
 	npm install --quiet
 
-dev: KG_PUBLIC_IP?=$$(docker-machine ip $$(docker-machine active))
+dev: KG_PUBLIC_IP?=localhost
 dev: kernel-gateway-container ## Run dashobard server on localhost
 	NODE_ENV='development' KG_BASE_URL=$(KG_BASE_URL) KERNEL_GATEWAY_URL=http://$(KG_PUBLIC_IP):$(KG_PUBLIC_PORT) gulp
 
-dev-logging: KG_PUBLIC_IP?=$$(docker-machine ip $$(docker-machine active))
+dev-logging: KG_PUBLIC_IP?=localhost
 dev-logging: kernel-gateway-container ## Run dashboard server with debug logging on localhost
 	NODE_ENV='development' gulp build
 	KERNEL_GATEWAY_URL=http://$(KG_PUBLIC_IP):$(KG_PUBLIC_PORT) npm run start-logging
 
-dev-debug: KG_PUBLIC_IP?=$$(docker-machine ip $$(docker-machine active))
+dev-debug: KG_PUBLIC_IP?=localhost
 dev-debug: kernel-gateway-container ## Run dashboard server with node debugger on localhost
 	NODE_ENV='development' KG_BASE_URL=$(KG_BASE_URL) KERNEL_GATEWAY_URL=http://$(KG_PUBLIC_IP):$(KG_PUBLIC_PORT) gulp debug
 
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,9 @@ HTTP_PORT?=3000
 HTTPS_PORT?=3001
 TEST_CONTAINER_NAME:=$(DASHBOARD_CONTAINER_NAME)-test
 TEST_IMAGE_NAME:=jupyter-incubator/$(TEST_CONTAINER_NAME)
+PROXY_CONTAINER_NAME:=reverse-proxy
+PROXY_IMAGE_NAME:=nginx:alpine
+PROXY_PUBLIC_PORT:=8080
 
 help:
 # http://marmelab.com/blog/2016/02/29/auto-documented-makefile.html
@@ -51,7 +54,7 @@ build: images
 
 define KILL
 @echo '-- Removing Docker containers'
-@-docker rm -f $(DASHBOARD_CONTAINER_NAME) $(KG_CONTAINER_NAME) $(TEST_CONTAINER_NAME) 2> /dev/null || true
+@-docker rm -f $(DASHBOARD_CONTAINER_NAME) $(KG_CONTAINER_NAME) $(TEST_CONTAINER_NAME) $(PROXY_CONTAINER_NAME) 2> /dev/null || true
 endef
 
 kill: ## Kill all running docker containers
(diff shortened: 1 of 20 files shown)


jhpedemonte:
@danielfrg, I started reviewing your PR, but ran into some issues. Then I ended up creating my own patch, based on yours but with some changes/fixes.

Main diffs from that PR:
* Used name `prefixUrl`, since `baseUrl` can have different connotation, depending on the context.
* Make sure to use `urljoin` instead of string concatenation when dealing with paths/urls.
* Cleaner impl of changes in the handlebar templates. I didn't like having `urlJoin` do the prefix magic. Instead, I pass in the `prefixUrl` variable for use in the correct places.
* Fix for font errors.

jhpedemonte:
I haven't tested everything yet (hence the "WIP" in the title). Will continue next week.

I did notice that I'm getting these errors when loading the **INTRO** notebook:

```
dashboard.js:59067 GET http://127.0.0.1:3000/components/b80c321cdb09746d881e6b9227fa4491.png 404 (Not Found)
```

danielfrg:
This is great. I had not a lot of love for my implementation and I am sure yours is better.

I am out this week but I'll try to find to to test if needed.

jhpedemonte:
Actually, I'm not thinking about this right. #260 is about the dashboard server residing behind a reverse proxy, where the client facing URL will have a prefix (i.e. http://something.com/dashboard-server/...). In that scenario, we really don't need to change how the node/express app serves its routes -- most of that stays the same. Instead, we would need to fix templates (already done), plus maybe some of the websocket proxying, etc.

In order to test this, we'll need to stand up a reverse-proxy in front of the dashboard server. Shouldn't be too hard using Docker.

danielfrg:
That makes sense for a reverse proxy but it won't be enough for a the `configurable-http-proxy` that JupyterHub uses. My final objective with my PR (and now yours) is to register the `jupyter-dashboard-server` as a service on JupyterHub that can be used with JupyterHub authentication.

jhpedemonte:
I'm aware of JupyterHub, just never set it up or played with the source. And `configurable-http-proxy` is new to me -- although it's just a wrapper to `node-http-proxy`, which we use.

Looks like `configurable-http-proxy` can be set up as a simple reverse proxy also, similar to what I [described above](https://github.com/jupyter-incubator/dashboards_server/pull/292#issuecomment-249341835). Through some combination of these arguments:

```
    --no-prepend-path                Avoid prepending target paths to proxied requests
    --no-include-prefix              Don't include the routing prefix in proxied requests
```

That way, you could add a route to `configurable-http-proxy` like so:

```
POST /api/routes/foo/bar -> http://somehost:1234
```

Then, calls to `http://config-http-proxy.com/foo/bar` would get routed to `http://somehost:1234` -- "prefix" route removed.

Is this sufficient for your needs?  I think it would be best to keep the code as simple as possible and not add the ability to run from a prefixed route if we don't need to.

parente:
> That way, you could add a route to configurable-http-proxy like so:

```
POST /api/routes/foo/bar -> http://somehost:1234
```

I think the issue is that it's not possible to tell JupyterHub to use the configurable-http-proxy this way. I may be wrong, but I think that JupyterHub relies on setting the `base_url` option for notebook servers so that it can write routes like:

```
/user/parente/ -> http://somehost:1234/user/parente/
```

danielfrg:
Yes, thats correct the Hub does rely in that notebook setting for the single user notebook (and JupyterLab) servers. I was trying to get the same functionality in here.

laurikoobas:
Looked through the proposal and changes, sounds exactly like the solution for the problem I had.

How can I gain access to the result? Doesn't look like I can clone the incubator repository in a way that it has this specific branch in it.

jhpedemonte:
@parente Well, that's unfortunate. So sounds like we'll need to have two config options:
1. `PROXIED` - Set to true to let the server know that it is behind a proxy. [Express must be told it is behind a proxy](https://expressjs.com/en/guide/behind-proxies.html). This value should also allow any of the values described in that page.
2. `BASE_PATH` - (or something like that) Defines the base prefix path that the server operates with.

For a classical reverse proxy situation, an admin would only have to specify `PROXIED` (I think; almost got that working). In the JupyterHub situation, admin would define both.  Does that seem legit?

jhpedemonte:
@laurikoobas This PR is still a work in progress. While it is mostly working, it is not fully working yet. That said, if you want to test the PR as it currently stands you can either clone my repo and branch (jhpedemonte/dashboards_server@260-prefix-url) or apply this [patch](https://github.com/jupyter-incubator/dashboards_server/pull/292.patch) to your clone of this repo.

parente:
@jhpedemonte The two options sound right to me, assuming running the dashboard server behind the jupyterhub proxy as a managed service makes sense. Maybe @minrk can chime in on whether that fits the model correctly.

jhpedemonte:
Updated patch with support for both passing and not passing the prefix value. Seems to be working all around. Looking at test issues now.

jhpedemonte:
Tests run fine on my machine, but failing on Travis. Yay!

@parente or @dalogsdon, would appreciate a review.  @danielfrg, please test this patch when you get a chance. I did not test on JupyterHub, just using Nginx.

danielfrg:
I am a little busy this week but I will try to test it as soon as I can. Thanks!

laurikoobas:
@jhpedemonte tried the patch and it's working. One addition I'd like is just the startup message to mention it as well:
`Jupyter dashboard server listening on 127.0.0.1:3000`
should be with the path added.

minrk:
Those options make sense for working JupyterHub. A Hub-managed service will get the JUPYTERHUB_SERVICE_PREFIX environment variable, which will be `/services/service-name` in most cases. It also receives JUPYTERHUB_BASE_URL, which is the base URL of the whole application (typically `/`, and always the part before `/services` in SERVICE_PREFIX).

jhpedemonte:
@laurikoobas Done.

@minrk Great, thanks for the info.

parente:
I'll try to kick the tires on this by the end of the week.

danielfrg:
I just tested this in a couple of settings and it looks like everything its working.

I tested multiple `prefixUrl` values including more "complex" ones like: `/hub/service/dashboards/server1`.

I also tested this behind the `configurable-http-proxy`, all good as far as I can see.

dalogsdon:
Taking a look now

dalogsdon:
Looks good to me. The only issue I saw is the lack of calling out the example nginx config file in the README. I had to read the Makefile to know where the example config is.

parente:
Code looks fine to me.

Maybe I missed it, but is there a unit/integration test running with the base URL set to something other than the default? Do we want to do that for one test at least? Not with the proxy, just with a `/test-root` base URL or something.

bowenli37:
Just tried to run proxy-container by "make proxy-container". Containers started normally. Visited http://localhost but received 404 error message. I noted that nginx container port 8080 is mapped to host port 80 in Makefile, but nginx is listening on port 80 as in conf file. After changing port 8080 to 80 and rebuilding all 3 container, http://localhost/dashboards_server/ can be reached, but styling was not applied due to missing http://localhost/css/style.css (Failed to load resource). 
I was on the latest commit (# 5cac00) by today. What was I missing? Please help! Thanks very much.

jhpedemonte:
I responded to @bowenli37 in #307.

PR #295, opened 2016-09-28
Sets body height to prevent "dropup" menus

jhpedemonte:
Fixes jupyter-incubator/dashboards#278

Since cells are absolutely positioned, they don't affect the size of their parent/ancestor elements. This meant that `body` was quite small. The code that shows the dropdown menus would check if there was enough space to show below the widget; if not, it would show the menu above. This check was always showing not enough space. This change will set body to always take up at least the full viewport height.

### DIFF
--- a/client/less/dashboard.less
+++ b/client/less/dashboard.less
@@ -9,6 +9,7 @@ html {
     box-sizing: inherit;
 }
 body {
+    min-height: 100vh;
     font-size: 14px;
     margin: 0;
 }


parente:
Does this regress the fix for the scrollbar appearing all the time and causing jitter?

https://github.com/jhpedemonte/dashboards_server/commit/6714ded9229cab176adde308f18fd7db34a80a91

jhpedemonte:
Nope. In that issue, the CSS rule was on `outer-dashboard` -- this caused the scrollbar to appear since there was a nav bar sibling to `outer-dashboard`. In this case, I added the rule to `body`.

PR #297, opened 2016-10-12
Add authentication from Auth0

danielfrg:
This PR adds a simple auth from Auth0.

### DIFF
--- a/app/auth-auth0.js
+++ b/app/auth-auth0.js
@@ -1,0 +1,54 @@
+/**
+ * Copyright (c) Jupyter Development Team.
+ * Distributed under the terms of the Modified BSD License.
+ */
+/**
+ * Example: Passport JS strategy for Auth0.
+ * Requires additional configuration and npm packages. See
+ * https://github.com/jupyter-incubator/dashboards_server/wiki/Authentication
+ * for details.
+ */
+var config = require('./config');
+var passport = require('passport');
+var Auth0Strategy = require('passport-auth0');
+
+module.exports = function(app) {
+
+    app.get('/callback',
+        passport.authenticate('auth0', { failureRedirect: '/login' }),
+        function(req, res) {
+            if (!req.user) {
+                throw new Error('user null');
+            }
+            res.redirect("/");
+        }
+    );
+
+    app.get('/login',
+        passport.authenticate('auth0', {}), function (req, res) {
+            res.redirect("/");
+    });
+
+    app.post('/logout', function(req, res){
+        req.logout();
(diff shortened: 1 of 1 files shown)


parente:
Thanks for the contribution. Can you add a note about it on the auth wiki page so folks know how to use it? https://github.com/jupyter-incubator/dashboards_server/wiki/Authentication

danielfrg:
I don't have permission to edit the wiki. Should I write it in a markdown file? I could move the auth docs from the wiki to the repo and add the auth0 specifics.

parente:
> I don't have permission to edit the wiki.

Oh wow. I thought the wiki used to be public read/write. I guess the default changed?

> I could move the auth docs from the wiki to the repo and add the auth0 specifics.

If you're feeling up to the task, that's very generous of you. Alternatively, you can drop the markdown here and I can do it. Or, if you're planning on contributing more over time, we can add you as a committer so you can update the wiki directly.

danielfrg:
I am not the best node dev but I am planning on using the dashboards more so that will probably include a couple of contributions. So if you can add me as a contributor I can add the docs to the wiki.

parente:
Invite in the mail.

app/auth-auth0.js
```diff
@@ -0,0 +1,54 @@
...
+var passport = require('passport');
+var Auth0Strategy = require('passport-auth0');
+
+module.exports = function(app) {
+
+    app.get('/callback',
+        passport.authenticate('auth0', { failureRedirect: '/login' }),
+        function(req, res) {
+            if (!req.user) {
+                throw new Error('user null');
```
jhpedemonte: Provide a more useful error message here, such as "User name must be set".

app/auth-auth0.js
```diff
@@ -0,0 +1,54 @@
...
+    );
+
+    app.get('/login',
+        passport.authenticate('auth0', {}), function (req, res) {
+            res.redirect("/");
+    });
+
+    app.post('/logout', function(req, res){
+        req.logout();
+        res.redirect('https://' + config.get('AUTH0_DOMAIN') + '/v2/logout?returnTo=' + config.get('PUBLIC_LINK_PATTERN') + '&client_id=' + config.get('AUTH0_CLIENT_ID'));
```
jhpedemonte: I think it's better to construct URLs using an API rather than string concatenation. In particular, this works around issues where a user may or may not append a '/'.

In other parts of the code, we've used `url-join`, but not sure how well it handles URL params. If that doesn't work, consider using Node.js' [`url.format()`](https://nodejs.org/dist/latest-v4.x/docs/api/url.html#url_url_format_urlobj).

danielfrg:
> Invite in the mail.

Thanks! I added a small section to the wiki on Auth0.

PR #303, opened 2016-11-12
Fix regression in title link nav

parente:
Handle the default case where baseUrl is empty. Otherwise, it just links back to the current page.


parente:
Rebased on master to fix test failure.

PR #304, opened 2016-11-12
Bump to 1.1.x kernel gateway

parente:


parente:
Travis failures are due to a race condition in the test code.

PR #305, opened 2016-11-14
Add missing admin script to package

parente:
Match the script name to the existing server script too


jhpedemonte:
So this ensures that admin script is added to the **npm** package?

parente:
I believe / hope so. It's missing in the current release on npm.

